PR #21, opened 2016-06-12
Allow using internal interfaces (e.g. "Traversable") as parameter types

### LINKED ISSUE #20: The "\Go\ParserReflection\ReflectionParameter::getClass" fails when parameter has internal interface type
The `\Go\ParserReflection\ReflectionParameter::getClass` method works fine with user-defined and internal classes. It uses own `\Go\ParserReflection\ReflectionClass` class) instead of `\ReflectionClass` when method parameter type hint is an internal interface, like `Traversable`.

aik099:
Closes #20 

P.S.

### DIFF
--- a/src/ReflectionParameter.php
+++ b/src/ReflectionParameter.php
@@ -184,9 +184,9 @@ class ReflectionParameter extends BaseReflectionParameter
                 throw new ReflectionException("Can not resolve a class name for parameter");
             }
             $className   = $parameterType->toString();
-            $classExists = class_exists($className, false);
+            $classOrInterfaceExists = class_exists($className, false) || interface_exists($className, false);
 
-            return $classExists ? new \ReflectionClass($className) : new ReflectionClass($className);
+            return $classOrInterfaceExists ? new \ReflectionClass($className) : new ReflectionClass($className);
         }
 
         return null;
--- a/tests/Stub/FileWithParameters.php
+++ b/tests/Stub/FileWithParameters.php
@@ -20,7 +20,8 @@ function miscParameters(
     &$byReferenceParam,
     &$byReferenceNullable = __CLASS__,
     $constParam = TEST_PARAMETER,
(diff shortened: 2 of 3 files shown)


src/ReflectionParameter.php
```diff
@@ -184,7 +184,7 @@ public function getClass()
                 throw new ReflectionException("Can not resolve a class name for parameter");
             }
             $className   = $parameterType->toString();
-            $classExists = class_exists($className, false);
+            $classExists = class_exists($className, false) || interface_exists($className, false);
```
lisachenko: Parameter should be renamed to $classOrInterfaceExists here

aik099: Done.

lisachenko:
It would be nice to have a test for that case.
Here is a guide, how to do this:
1) Open `FileWithParameters` file and add into `miscParameters` function one more argument with typehint (better to the end of args):
```php
    \Traversable $traversable
```
2) Add a test for this into the `ReflectionParameterTest` into the `testGetClassMethod()` test case:
```php
        $traversableInterface = $parameters[12 /* \Traversable $traversable */]->getClass();
        $this->assertInstanceOf(\ReflectionClass::class, $traversableInterface);
        $this->assertSame(\Traversable::class, $traversableInterface->getName());
```

That's it, now you will have a test case for that

aik099:
Added tests and squashed.

tests/Stub/FileWithParameters.php
```diff
@@ -20,7 +20,8 @@ function miscParameters(
     &$byReferenceParam,
     &$byReferenceNullable = __CLASS__,
     $constParam = TEST_PARAMETER,
-    $constValueParam = __NAMESPACE__ // This line is long and should be truncated
+    $constValueParam = __NAMESPACE__, // This line is long and should be truncated
```
aik099: Not sure why CS reports an issue here.

lisachenko:
Thanks! Merged

PR #22, opened 2016-06-12
Limit number of cached files (with a setting)

### LINKED ISSUE #18: Helper methods for avoiding memory leaks
The "\Go\ParserReflection\ReflectionEngine::parseClass" method is caching each processed file in `\Go\ParserReflection\ReflectionEngine::$parsedFiles` property. This is correct behavior and allows to avoid high CPU load, when parsing code base, when parsing classes, that extend few common classes.

However this behavior also leads to memory leak, when parsing lots of unrelated classes, because each of classes is parsed only once and after being parsed is never reused.

Proposing to add following methods, that would allow user to manually remove cached parsed file:

```php
/**
 * Removes parsing result of a file
 *
 * @param string      $fileName Name of the file
 *
 * @return void
 * @throws \InvalidArgumentException When file isn't parsed.
 */
public static function unsetFile($fileName)
{
    if (!isset(self::$parsedFiles[$fileName])) {
        throw new \InvalidArgumentException('The "' . $fileName . '" isn\'t parsed.');
    }

    unset(self::$parsedFiles[$fileName]);
}
```

aik099:
I've decided to clear all parsed files, when limit is reached instead just removing last parsed file. This technique is looking more promising for cases, when you parse several large files upfront and can't lower memory usage, because they can't be removed later.

Closes #18

### DIFF
--- a/src/ReflectionEngine.php
+++ b/src/ReflectionEngine.php
@@ -37,6 +37,11 @@ class ReflectionEngine
     protected static $parsedFiles = array();
 
     /**
+     * @var null|integer
+     */
+    protected static $maximumCachedFiles;
+
+    /**
      * @var null|Parser
      */
     protected static $parser = null;
@@ -64,6 +69,18 @@ class ReflectionEngine
         $traverser->addVisitor(new NameResolver());
 
         self::$locator = $locator;
+    }
+
+    /**
+     * Limits number of files, that can be cached at any given moment
+     *
+     * @param integer $newLimit New limit
+     *
+     * @return void
+     */
+    public static function setMaximumCachedFiles($newLimit)
+    {
+        self::$maximumCachedFiles = $newLimit;
     }
 
     /**
@@ -186,6 +203,10 @@ class ReflectionEngine
     {
         if (isset(self::$parsedFiles[$fileName])) {
             return self::$parsedFiles[$fileName];
+        }
+
+        if (isset(self::$maximumCachedFiles) && (count(self::$parsedFiles) === self::$maximumCachedFiles)) {
+            array_shift(self::$parsedFiles);
         }
 
         if (!isset($fileContent)) {


aik099:
While trying how this approach works I've found that it's getting worse, when you're analyzing 100 classes that have same parent. When you set limit to 10 classes, then you have to parse that parent class 10 times (once per each 10 parsed classes).

Any ideas?

aik099:
I've tried 3 ways:

* clearing parsed files array
* removing 1st element (as you suggested)
* removing last element

No matter what I try it's pretty slow (I guess because what gets removed is needed again and again). When using manual unsetting the results are drastically better.

lisachenko:
I still think that my way with configurable number of items is still most suited for this needs.
Clearing all parsed files is not good, because there will be a lot of misses and engine will be doing extra parsing each time.
Removing last element is not good too, because it can be just added item or interface and chances that it will be used one more time is greater than for first (oldest one) element.

If you notice that performance is not good for you, then you should increase the number of elements to 100 or even 1000. Anyway, you can add more memory for PHP, more items in the cache, etc. No easy way to store all items directly in the cache for big project, we should process classes iteratively.

I'm going to accept a PR with logic of cleaning 1st element in the cache. All other ways can be worse and definitely can slow down the parsing process.

aik099:
> I'm going to accept a PR with logic of cleaning 1st element in the cache. All other ways can be worse and definitely can slow down the parsing process.

Will do. Maybe we can not just remove oldest cached element, but instead remove elements with less hits? Goal is to keep N parent classes/interfaces in cache, that everybody uses. This way one-time classes will be quickly removed.

> Also, please, could you make the API for setting this limit? (not with public static property)

What is API setting? There are no configuration examples currently or configuration at all. Do you mean `ReflectionEngine::init` method parameter?

lisachenko:
> What is API setting? There are no configuration examples currently or configuration at all. Do you mean ReflectionEngine::init method parameter?

I think it would be better to have a static setter, something like `setMaximumCachedFiles($newLimit)` for `ReflectionEngine` class

aik099:
> I think it would be better to have a static setter, something like setMaximumCachedFiles($newLimit) for ReflectionEngine class

Will do.

aik099:
Made requested changes.

aik099:
What you thinking about idea from https://github.com/goaop/parser-reflection/pull/22#issuecomment-225555286 about removing elements with less hits? Just removing oldest element (if that element is sub-class of every class) would cause it's being reparsed. If we increase allowed file count in cache, then it would increase used memory amount which will defeat the purpose of entire change.

lisachenko:
PR is ready for merge.

> What you thinking about idea from #22 (comment) about removing elements with less hits?

I think, that this idea is good, but the nature of classes isn't equal to the nature of content that can be cached. This means that no "hit-rate" for classes for a typical PHP project, I'm expect class-loading and analysing process to be linear, so all new classes are loaded during the request processing, there is a little chance that some previous class will be reused one more time.

Anyway, we can control hit-rate for our cache by increasing the memory and by increasing the limit for stored files. So, I'm confident that it will be enough.

If you agree with me, then ping me with merge.

aik099:
OK. Let's merge this in then and see how it will turn out.

The best test case would be using it on Symfony code base to see what memory limit and file count in cache is best for performance.

lisachenko:
Merged

lisachenko:
Do you want to do something more in the library for this version? If it's ok for you now, then I can release version 1.1.0 with all of your PRs.

aik099:
Maybe a few bugfixes (if I find any), but for now things look stable. You can release new feature version safely.

PR #24, opened 2016-06-12
Support "self" and "parent" as method parameter type hint

### LINKED ISSUE #23: Fatal error, when using "self" or "parent" as method parameter type hint
Not sure why, but PHP-Parser doesn't resolve `self` or `parent` to actual FQCN when used as class method parameter type hint. That results in exception with `Can not resolve a class name for parameter` message being thrown.

The `\Go\ParserReflection\ValueResolver\NodeExpressionResolver::fetchReflectionClass` should have taken care about that, but it didn't or it's not working for parameter type hints, but it's used only for resolving default value of parameter.

aik099:
Closes #23

P.S.

### DIFF
--- a/tests/Stub/FileWithParameters.php
+++ b/tests/Stub/FileWithParameters.php
@@ -30,3 +30,7 @@ class Foo {
     public function methodParam($firstParam, $optionalParam = null) {}
     public function methodParamConst($firstParam = self::CLASS_CONST, $another = __CLASS__, $ns = TEST_PARAMETER) {}
 }
+
+class SubFoo extends Foo {
+    public function anotherMethodParam(self $selfParam, parent $parentParam) {}
+}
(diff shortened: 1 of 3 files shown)


lisachenko:
Could you please, add a test? Example is described in previous PR.

aik099:
> Could you please, add a test? Example is described in previous PR.

Will do.

aik099:
Added tests and squashed.

tests/Stub/FileWithParameters.php
```diff
@@ -30,3 +30,7 @@ class Foo {
     public function methodParam($firstParam, $optionalParam = null) {}
     public function methodParamConst($firstParam = self::CLASS_CONST, $another = __CLASS__, $ns = TEST_PARAMETER) {}
 }
+
+class SubFoo extends Foo {
```
aik099: Not sure if this needs to be fixed, because these are just test fixtures.

tests/Stub/FileWithParameters.php
```diff
@@ -30,3 +30,7 @@ class Foo {
...
+    public function methodParam(self $selfParam, parent $parentParam) {}
```
aik099: Not sure if this needs to be fixed, because these are just test fixtures.

lisachenko:
Merged

PR #25, opened 2016-06-12
Fixing notice, when using "ReflectionClass::hasConstant" method

aik099:
When checking class constant has `NULL` as it's value and it's being checked using `ReflectionClass::hasConstant` method then there will be a notice and constant will be returned as non-existent.

This happens, because `array_key_exists` function has different argument order.

Most likely tests doesn't cover this use case. I've found this code in Symfony framework.

P.S.


lisachenko:
Thanks! Merged, test would be nice to have.

PR #26, opened 2016-06-13
Add tests for "ReflectionClass::hasConstant" and "ReflectionClass::getConstant" methods

aik099:
While adding tests for `ReflectionClass::hasConstant`, that was fixed in #25 I've also found out, that `ReflectionClass::getConstant` isn't directly tested as well. So I've added tests for it as well.

Related to #25


lisachenko:
Thanks!

PR #27, opened 2016-06-13
Adding "CallableLocator" class locator

### LINKED ISSUE #19: Support for non-Composer-based class locator
In a project I'm analyzing 2 autoloaders are being used in combination. To cover that case I've created `CallableLoader`, that accepts any callback in constructor and will call it to locate class file.

If you're ok I can submit a PR with it.

aik099:
Closes #19


lisachenko:
Thanks! Merged

PR #28, opened 2016-06-14
Adding test for "ComposerLocator"

aik099:
The `realpath` call is needed to resolve relative path, that Composer autoloader is building since PHP 5.6. The related Composer feature is called static autoloader.

### DIFF
--- a/tests/Locator/ComposerLocatorTest.php
+++ b/tests/Locator/ComposerLocatorTest.php
@@ -1,0 +1,19 @@
+<?php
+namespace Go\ParserReflection\Locator;
+
+use Go\ParserReflection\ReflectionClass;
+
+class ComposerLocatorTest extends \PHPUnit_Framework_TestCase
+{
+    public function testLocateClass()
+    {
+        $locator = new ComposerLocator();
+
+        $reflectionClass = new \ReflectionClass(ReflectionClass::class);
+
+        $this->assertSame(
+            $reflectionClass->getFileName(),
+            realpath($locator->locateClass(ReflectionClass::class))
+        );
+    }
+}


lisachenko:
Hm, nice catch with this new feature :+1: 

Unfortunately, `realpath` function can't be used for phar archives and stream wrappers. Looks like https://github.com/goaop/framework/blob/master/src/Instrument/PathResolver.php implementation should be copied for that.

aik099:
But we're using `realpath` call in test suite. Are you suggesting that test suite can also be run when this library is compiled into PHAR and that's why this PR can't be merged?

lisachenko:
This PR can't be merged because it doesn't fix the issue, we should check that locator returns normalized path. You have added a fix in the test itself (by adding `realpath`), but `ReflectionClass->getFileName()` method now returns incorrect path and test can be broken.

aik099:
Yes, if we look at `ComposerLocator` class alone, then it indeed it returns different path for different PHP versions, but since it's used only internally it doesn't break anything.

The locator class is not used to power result of `\Go\ParserReflection\ReflectionClass::getFileName` method. Actual path, that is returned by `\Go\ParserReflection\ReflectionClass::getFileName` method comes from PHP-Parser, which does `realpath` internally.

lisachenko:
> The locator class is not used to power result of \Go\ParserReflection\ReflectionClass::getFileName method. Actual path, that is returned by \Go\ParserReflection\ReflectionClass::getFileName method comes from PHP-Parser, which does realpath internally.

It's not correct.See `Go\ParserReflection\ReflectionEngine::parseClass` method:

```php
public static function parseClass($fullClassName) 
{
    $classFileName  = self::locateClassFile($fullClassName);
    //...
    if ($namespaceLevelNode instanceof ClassLike && $namespaceLevelNode->name == $className) {
            $namespaceLevelNode->setAttribute('fileName', $classFileName);
    }
}
```

And then, `Go\ParserReflection\Traits\ReflectionClassLikeTrait::getFileName` returns this value:
```php
    public function getFileName()
    {
        return $this->classLikeNode->getAttribute('fileName');
    }
```

aik099:
I see now. Apparently tests covering `getFileName` is missing. I'll add one as part of this PR and then update `ComposerLocatorTest`.

aik099:
And it seems, that following methods needs to call `stream_resolve_include_path` upon given filename, because they are not only used internally by the library, but also can be used manually:

* `ReflectionFileNamespace::__construct`
* `ReflectionFile::__construct`
* `ReflectionEngine::parseFile`

lisachenko:
Correct way will be to take a `PathResolver` class from goaop framework for resolving paths, because I want to keep this functionality for phar streams too. And then use it for every place that require normalization.

`stream_resolve_include_path` can't be used for custom streams too, so if `__DIR__` for composer will be equal to something like this: `phar://package.phar/vendor/composer` then nothing will work...

aik099:
We might use https://github.com/box-project/box2-path instead. It's already packaged as reusable library and is quite minimalistic.

aik099:
> stream_resolve_include_path can't be used for custom streams too, so if __DIR__ for composer will be equal to something like this: phar://package.phar/vendor/composer then nothing will work...

Are you suggesting people would create custom stream wrapper (e.g. `s3://`) and try to analyze files in there? Not sure if anybody would do that in real life. It's documentation (see http://php.net/manual/en/function.stream-resolve-include-path.php) doesn't say that it won't work with custom stream wrappers. Since it works the same way as path resolver used internally for `fopen` then it's pretty good for us, nope?

lisachenko:
Package box2-path is not standalone, it depends on `phine/exception` package. I'm not so happy to include additional dependencies for small functions (do you remember the case with npm's left-pad packet?) 

Also, I'm not happy to copy-paste the same code twice..
However, support for stream wrappers is required for me, because goaop framework supports read-only file systems and custom stream wrappers. Previously, composer returns nice paths without dots and even phar:// classes were analysed easily.

aik099:
> Package box2-path is not standalone, it depends on phine/exception package. I'm not so happy to include additional dependencies for small functions (do you remember the case with npm's left-pad packet?)

Nope, but I think I get an idea. You depend on small package and that package depends on a lot of packages by itself.

> Also, I'm not happy to copy-paste the same code twice..
However, support for stream wrappers is required for me, because goaop framework supports read-only file systems and custom stream wrappers.

I'm happy for `goaop framework`, but how this library is related to all of that?

> Previously, composer returns nice paths without dots and even phar:// classes were analysed easily.

Why it should stop working now. The `phar://path/to/some/folder/../another-folder/file.php` should working currently as well.

P.S.

Still I think we should be taking this discussion as part of separate issue. I've only added missing test here and that test is fine by itself. Fact, that all functions working with file paths turns out were expecting absolute path and this wasn't covered by tests is absolutely unrelated issue. I've created #31 about that.

lisachenko:
> I've only added missing test here and that test is fine by itself.

Test is not correct, because it has a `realpath` that hides the issue #31. I can merge this with additional requirement to remove this normalization in future version.

aik099:
> https://github.com/goaop/parser-reflection/pull/28#issuecomment-225872647

What about this? You haven't answered anything.

aik099:
> Test is not correct, because it has a realpath that hides the issue #31.

I guess you expected locator class to return absolute path. I would gladly fix that issue in `ComposerLocator` class once we done on faith of `stream_resolve_include_path`.

lisachenko:
> What about this? You haven't answered anything.

This won't work. See https://bugs.php.net/bug.php?id=52769

aik099:
> This won't work. See https://bugs.php.net/bug.php?id=52769

Question in there is totally not about `realpath` 😄  I've asked about usage of `__DIR__` and `stream_resolve_include_path`. You seem to belive that it won't work with custom stream handlers, but function documentation doesn't say that exactly.

lisachenko:
> You seem to belive that it won't work with custom stream handlers, but function documentation doesn't say that exactly.

```php
$phar = new Phar('test.phar');
$phar['index.php'] = <<<'FILE'
<?php
$parts   = explode(DIRECTORY_SEPARATOR, __DIR__);
$lastDir = array_pop($parts);
var_dump(__FILE__, stream_resolve_include_path(__DIR__.'/../' . $lastDir . '/index.php'));
FILE;


$parts   = explode(DIRECTORY_SEPARATOR, __DIR__);
$lastDir = array_pop($parts);
var_dump(__FILE__, stream_resolve_include_path(__DIR__.'/../' . $lastDir . '/index.php'));
include 'phar://test.phar';
```

Shows:
```
./test.php:14:string '<DIR>/index.php' (length=26)
./test.php:14:string '<DIR>/index.php' (length=26)

phar://<DIR>/test.phar/index.php:4:string 'phar://<DIR>/test.phar/index.php' (length=43)
phar://<DIR>/test.phar/index.php:4:boolean false
```

Pay an attention to the last line, it's output from the phar archive and you can see, that it didn't work. So, please trust me, because I spent a lot of time when with normalization of paths for streams and archive.

aik099:
Any chance of moving `PathResolver` class into separate repo/package?

lisachenko:
> Any chance of moving PathResolver class into separate repo/package?

It's too small to be extracted into a small package... Maybe just copy it and deal with it?

aik099:
So we:

* copy it to a project
* change namespace to match
* as part of this PR use it in ComposerLocator
* as part of PR for #31 use it for all methods mentioned in there

?

lisachenko:
> copy it to a project, change namespace to match, use it for all methods mentioned in there

Yes, as a part for the #31 PR
After that, we can proceed with this PR by removing a `realpath` from test and applying it on top of #31

PR #29, opened 2016-06-14
Adding "branch-alias" to "composer.json"

aik099:
When `branch-alias` present in `composer.json` it comes possible to specify version like `^1.1@dev` when requiring the library. This way project using the library will get it's dev dependency, but whoever uses that project would use live dependency instead.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,5 +27,10 @@
   },
   "require-dev": {
     "phpunit/phpunit": "~4.0"
+  },
+  "extra": {
+    "branch-alias": {
+      "dev-master": "1.x-dev"
+    }
   }
 }


composer.json
```diff
@@ -27,5 +27,10 @@
   },
   "require-dev": {
     "phpunit/phpunit": "~4.0"
+  },
+  "extra": {
+    "branch-alias": {
+      "dev-master": "1.1.x-dev"
```
lisachenko: better to use 1.x-dev here to avoid frequent changes...

aik099: Done.

lisachenko:
Merged

PR #36, opened 2016-06-23
Resize the cache size after changing the maximum limti, resolves #35

### LINKED ISSUE #35: The "ReflectionEngine::setMaximumCachedFiles" doesn't reevaluate existing cache
When `ReflectionEngine::setMaximumCachedFiles` method is called it currently only changes maximal allowed file count in cache. That's perfectly fine, when used prior to file parsing process is started.

Unfortunately, when used in the middle of process (e.g. when got high memory consumption I try to reduce cached file count), then nothing happens.

I expect existing cache to be sliced (from oldest to newest) to reduce cache to given size (or smaller if less files are cached, then requested).

lisachenko:


src/ReflectionEngine.php
```diff
@@ -81,6 +81,9 @@ public static function init(LocatorInterface $locator)
     public static function setMaximumCachedFiles($newLimit)
     {
         self::$maximumCachedFiles = $newLimit;
+        if (count(self::$parsedFiles) > $newLimit) {
+            self::$parsedFiles = array_slice(self::$parsedFiles, 0, $newLimit);
```
aik099: This would keep oldest entries in array. But when parsing new file oldest entry gets removed.

Maybe we should be slicing from right array side to be consistent?

aik099:
GitHub says that you've merged PR right after creating it. The point of PR is to allow others to review it before being merged. I would suggest waiting at least some amount of time before merging it.

lisachenko:
> GitHub says that you've merged PR right after creating it. The point of PR is to allow others to review it before being merged.

Yes, I've decided to merge it, because it's very simple. For bigger PR, I definitely like when someone can look at PR. 

Slicing an array from the right side can be good, but I don't see a big point in it.

aik099:
> Slicing an array from the right side can be good, but I don't see a big point in it.

To be consistent between `parseFile` and `setMaximumCachedFiles`. For example I'm expecting cache content to be identical for both following scenarios.

## Scenario No. 1

* call `parseFile(...)` 15 times
* call `setMaximumCachedFiles(10)`


## Scenario No. 2

* call `setMaximumCachedFiles(10)`
* call `parseFile(...)` 15 times

Unfortunately in current implementation `setMaximumCachedFiles` is cutting from the right (incorrect) , but `parseFile` is cutting from the left (correct). That would be a surprise to developer using `setMaximumCachedFiles` method in various places during parsing process and expecting a particular cache content after it's size was reduced.

PR #37, opened 2016-09-03
bugfix

wirwolf:
Fatal error: Call to undefined method PhpParser\Node\Stmt\ClassMethod::getParams() in /vendor/goaop/parser-reflection/src/Traits/ReflectionFunctionLikeTrait.php on line 157

### DIFF
--- a/src/Traits/ReflectionFunctionLikeTrait.php
+++ b/src/Traits/ReflectionFunctionLikeTrait.php
@@ -154,7 +154,7 @@ trait ReflectionFunctionLikeTrait
         if (!isset($this->parameters)) {
             $parameters = [];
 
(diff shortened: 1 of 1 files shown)


aik099:
Please add a test, that would cover fixed behavior.

aik099:
Actually this might not be a bug, but rather BC break in PHP-Parser dependency, that in 2.x version removed `getParams` method.

lisachenko:
Hi! Sorry for the late response, was lack of free time for my OSS projects. 

Just checked this issue and can't see anything wrong with `getParams()` method in the Php-parser 2.x branch. All 2.x tags also have this method.

So, looks like a false positive. Feel free to reopen if it can be reproduced.

PR #44, opened 2016-12-06
Fixed ReflectionParameters with class constant as default value

Bob4ever:


src/ValueResolver/NodeExpressionResolver.php
```diff
@@ -263,6 +263,9 @@ protected function resolveExprClassConstFetch(Expr\ClassConstFetch $node)
         if ('class' === $constantName) {
             return $refClass->getName();
         }
+        
+        $this->isConstant = true;
+        $this->constantName = (string)$node->class . '::' . $constantName;
```
aik099: What if this isn't constant from same class (e.g. `self::CONST_NAME`), but constant from other class (e.g. `OtherClass::CONST_NAME`) or even namespaced constant (declared via `const ...`)?

Would this work and is this covered in test case?

Bob4ever:
The behavior for namespace constants seems a little bit weird but it's the same with the default reflection api: https://3v4l.org/FqgbV

lisachenko:
Looks good for me. Squashed and merged, thanks!

Issue #45, opened 2016-12-06
Issues with PHP 5.6's class constant expressions


pdelre:
I believe I've found a couple issues with support for PHP 5.6's class constant expressions (http://php.net/manual/en/language.oop5.constants.php#example-179).

The class below will cause a segfault as parsing goes into an infinite loop:

```php
class Foo {
    const MINUTE = 60;
    const HOUR = 3600;

    const DEFAULT = self::MINUTE;
}
```

I am able to create other errors by adding the following class to `tests/Stubs/FileWithClasses56.php` (pulled from php.net's page above).

```php
const ONE = 1

class ClassWithConstantExpressions
{
    // below constants cause assertion failures, but no segfaults
    const TWO = ONE * 2;
    const THREE = ONE + self::TWO;
    const SENTENCE = 'The value of THREE is '.self::THREE;

    // uncommenting below will cause a segfault
    // const DEFAULT = self::TWO;
}
```

Failing output:
```
PHPUnit 4.8.30 by Sebastian Bergmann and contributors.

...............................................................  63 / 148 ( 42%)
II.....................F....................................... 126 / 148 ( 85%)
......................

Time: 157 ms, Memory: 10.00MB

There was 1 failure:

1) Go\ParserReflection\ReflectionClassTest::testGettersPHP56
getConstants() for class Go\ParserReflection\Stub\ClassWithConstantExpressions should be equal
Failed asserting that Array &0 (
    'TWO' => null
    'THREE' => null
    'SENTENCE' => null
) is identical to Array &0 (
    'TWO' => 2
    'THREE' => 3
    'SENTENCE' => 'The value of THREE is 3'
).

/Users/pdelre/Projects/github/parser-reflection/tests/ReflectionClassTest.php:299
/Users/pdelre/Projects/github/parser-reflection/tests/ReflectionClassTest.php:175

FAILURES!
Tests: 148, Assertions: 1714, Failures: 1, Incomplete: 2.
```

segfault failure:
```
PHPUnit 4.8.30 by Sebastian Bergmann and contributors.

...............................................................  63 / 148 ( 42%)
II.....................[1]    45899 segmentation fault  vendor/bin/phpunit tests
```

aik099:
You have shown 2 test runs. When does the segmentation fault happen?

pdelre:
When the `ClassWithConstantExpressions::DEFAULT` is uncommented.

```php
const ONE = 1

class ClassWithConstantExpressions
{
[...]

PR #46, opened 2016-12-06
Parse top level “define” statements in a root namespace

### LINKED ISSUE #32: Collect constants in a file defined via the "define" function
The `\Go\ParserReflection\ReflectionFileNamespace::getConstants` - method would return constants defined in a namespace (e.g. `const NAME = 'value';`) and not ones defined in a file defined in a file (e.g. `define('CONSTANT_NAME', 'value')`).

The `\Go\ParserReflection\ReflectionFileNamespace::getFunctions` method on the other hand properly returns functions defined in a file.

aik099:
Closes #32

### DIFF
--- a/src/ReflectionFileNamespace.php
+++ b/src/ReflectionFileNamespace.php
@@ -12,6 +12,8 @@ namespace Go\ParserReflection;
 
 use Go\ParserReflection\Instrument\PathResolver;
 use Go\ParserReflection\ValueResolver\NodeExpressionResolver;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Name;
 use PhpParser\Node\Stmt\ClassLike;
 use PhpParser\Node\Stmt\Const_;
 use PhpParser\Node\Stmt\Function_;
@@ -43,6 +45,13 @@ class ReflectionFileNamespace
      * @var array
      */
     protected $fileConstants;
+
+    /**
+     * List of constants in the namespace including defined via "define(...)"
+     *
+     * @var array
+     */
+    protected $fileConstantsWithDefined;
 
     /**
      * List of imported namespaces (aliases)
@@ -131,10 +140,20 @@ class ReflectionFileNamespace
     /**
      * Returns a list of defined constants in the namespace
      *
+     * @param bool $withDefined Include constants defined via "define(...)" in results.
+     *
      * @return array
      */
-    public function getConstants()
+    public function getConstants($withDefined = false)
     {
+        if ($withDefined) {
+            if (!isset($this->fileConstantsWithDefined)) {
+                $this->fileConstantsWithDefined = $this->findConstants(true);
+            }
+
+            return $this->fileConstantsWithDefined;
+        }
+
         if (!isset($this->fileConstants)) {
             $this->fileConstants = $this->findConstants();
         }
@@ -337,9 +356,11 @@ class ReflectionFileNamespace
     /**
      * Searches for constants in the given AST
      *
+     * @param bool $withDefined Include constants defined via "define(...)" in results.
+     *
      * @return array
      */
-    private function findConstants()
+    private function findConstants($withDefined = false)
     {
         $constants        = array();
         $expressionSolver = new NodeExpressionResolver($this);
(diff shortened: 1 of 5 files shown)


src/ValueResolver/NodeExpressionResolver.php
```diff
@@ -232,7 +232,7 @@ protected function resolveExprConstFetch(Expr\ConstFetch $node)
         if (!$isFQNConstant) {
             if (method_exists($this->context, 'getFileName')) {
                 $fileName      = $this->context->getFileName();
-                $namespaceName = $this->context->getNamespaceName();
+                $namespaceName = $this->resolveScalarMagicConstNamespace();
```
aik099: When `$this->context` is `ReflectionFileNamespace` class which has `getName` instead of `getNamespaceName` method original code (before this PR) ends up with fatal error about non-existing method call.

Luckily there is `resolveScalarMagicConstNamespace` method (not used anywhere except this PR), that specially handles that case.

src/ReflectionFileNamespace.php
```diff
@@ -357,6 +362,20 @@ private function findConstants()
             }
         }
 
+        if ($withDefinedConstants) {
+            foreach ($this->namespaceNode->stmts as $namespaceLevelNode) {
```
lisachenko: I can guess that we need to perform recursive scan here, not just a quick top-list nodes search. Example:

```php
if (!defined('SOME_CONST')) {
    define('SOME_CONST', 42);
}
```

So, maybe we initialize a node visitor here and use it for traversing, instead of direct scan?

src/ReflectionFileNamespace.php
```diff
@@ -357,6 +362,20 @@ private function findConstants()
...
+                if ($namespaceLevelNode instanceof FuncCall && (string)$namespaceLevelNode->name === 'define') {
```
lisachenko: You can not cast name to the string, because according to the BNF, name can be instance of `PhpParser\Node\Expr` and it doesn't contain `__toString()` method. So, it will be better to check that our argument can be cast to the string and then collect it's name.

I can suggest to implement this logic in the node visitor for collecting defines. See classes in the `Go\ParserReflection\NodeVisitor` namespace for examples.

src/ValueResolver/NodeExpressionResolver.php
lisachenko: Method `resolveScalarMagicConstNamespace()` is used dynamically for nodes of `\PhpParser\Node\Scalar\MagicConst\Namespace_` class (that holds \_\_NAMESPACE__ magic const).

However, I can agree that we can reuse it for our needs. So, :+1:

src/ValueResolver/NodeExpressionResolver.php
```diff
@@ -323,7 +323,7 @@ private function fetchReflectionClass(Node\Name $node)
         if (method_exists($this->context, 'getFileName')) {
             /** @var ReflectionFileNamespace|null $fileNamespace */
             $fileName      = $this->context->getFileName();
-            $namespaceName = $this->context->getNamespaceName();
+            $namespaceName = $this->resolveScalarMagicConstNamespace();
```
lisachenko: :+1:

src/ReflectionFileNamespace.php
```diff
@@ -131,12 +132,14 @@ public function getConstant($constantName)
     /**
      * Returns a list of defined constants in the namespace
      *
+     * @param bool $withDefinedConstants Include constants defined via "define(...)" in results.
+     *
      * @return array
      */
-    public function getConstants()
+    public function getConstants($withDefinedConstants = false)
```
lisachenko: You should separate caches with global constants by storing them in two different arrays in order to make it working for both cases and subsequent calls.

lisachenko:
Thank you for the PR! Review is in progress..

src/ReflectionFileNamespace.php
lisachenko: Ok, we can start with top-level defines only.

tests/ReflectionFileNamespaceTest.php
```diff
@@ -51,6 +52,28 @@ public function testGetConstant()
...
+
+        $this->assertSame(
+            array(
+                // Scalar types are handled.
+                'INT_CONST' => 5,
+                'STRING_CONST' => 'text',
+                'BOOLEAN_CONST' => true,
+
+                // Expressions are not handled.
+                'EXPRESSION_CONST' => null,
```
lisachenko: It's related to the #45, probably I'll add support soon for that.

src/ReflectionFileNamespace.php
aik099: What PHP code should like to have `FuncCall` class `name` property to be instance of `Expr` instead of `Name`?

aik099: What PHP code should look like to have `$namespaceLevelNode->name` property of `Expr` class instead of `Name` class?

src/ReflectionFileNamespace.php
aik099: Done.

src/ReflectionFileNamespace.php
aik099: Done. The code that caused this looks like `$define('TEST', 5);`.

tests/ReflectionFileNamespaceTest.php
aik099: Updated tests to reflect improved resolver capabilities.

lisachenko:
Squashed and merged. Thanks!

aik099:
Thanks, sorry for not being able to send PR sooner. There were a lot of other things at work, that needed my attention.

PR #50, opened 2016-12-07
Fix #48 NodeExpressionResolver will attempt to Reflect system classes

### LINKED ISSUE #48: NodeExpressionResolver will attempt to Reflect system classes
There are cases where the `NodeExpressionResolver` is attempting to Reflect system classes.

For example, the following will fail tests with an `InvalidArgumentException(Class DateTime was not found by locator)` from `ReflectionEngine::locateClassFile`:

Stub/FileWithClasses56.php

```php
class ClassWithComplexConstantsAndInheritance extends ClassWithArrayConstants
{
    const K = array(1, NS_CONST56);
    const L = [self::class, ClassWithArrayConstants::A, parent::B];
    // addition
    const M = \DateTime::ATOM;
}
```

I've seen this occur with parameters in production code, but haven't been able to reproduce it in the test suite yet.

pdelre:
Fixes #48

### DIFF
--- a/src/ValueResolver/NodeExpressionResolver.php
+++ b/src/ValueResolver/NodeExpressionResolver.php
@@ -446,9 +446,9 @@ class NodeExpressionResolver
         if ($isFQNClass) {
             // check to see if the class is already loaded and is safe to use
             // PHP's ReflectionClass to determine if the class is user defined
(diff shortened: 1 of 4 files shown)


lisachenko:
Thank you for the PR! Squashed and merged

PR #66, opened 2017-01-29
Support for Composer classmap autoloading

hijarian:
In the classmap autoloader mode of Composer, class names are written without the leading slash, but `ComposerLocator` passes them with the leading slash.

This leads to the following error when I try to use, for example, AspectMock library on Yii 1.1-based PHP project with PHP version 7.1 (see screenshot):

Feel free to ask me for more details.

All 2000+ tests are still passing and I believe this change actually fixes a long-standing problem instead of creating another one.

### DIFF
--- a/src/Locator/ComposerLocator.php
+++ b/src/Locator/ComposerLocator.php
@@ -51,7 +51,7 @@ class ComposerLocator implements LocatorInterface
      */
     public function locateClass($className)
     {
-        $filePath = $this->loader->findFile($className);
+        $filePath = $this->loader->findFile(ltrim($className, '\\'));
         if (!empty($filePath)) {
             $filePath = PathResolver::realpath($filePath);
         }


aik099:
> In the classmap autoloader mode of Composer, class names are written without the leading slash, but ComposerLocator passes them with the leading slash.

How Composer is able to understand both classmap and psr-4/psr-0 resolves class names then?

hijarian:
I'm removing this pull request, can't reproduce the original problem even with 1.2.0 version anymore.

hijarian:
@lisachenko @aik099 Actually, after reinstalling the Composer dependencies problem reappeared so I returned to this fix but I believe that https://github.com/goaop/parser-reflection/pull/67 fixes it in more appropriate way so I'll not reopen this PR.

hijarian:
Here's minimal setup to reproduce the problem:

`composer.json`
```
{
    "require": {
        "goaop/parser-reflection": "^1.2",
    },
    "autoload": {
        "classmap": [
            "src"
        ]
    }
}
```

`src\MyObjectType.php`:
```
<?php

class MyObjectType
{

}
```

`main.php`
```
<?php
require_once("vendor/autoload.php");

$locator = new \Go\ParserReflection\Locator\ComposerLocator();

var_dump($locator->locateClass('\MyObjectType')); // bool(false)
var_dump($locator->locateClass('MyObjectType')); // string(51) "L:\projects\goaop-parser-pr-66\src\MyObjectType.php"

new MyObjectType; // no exception, autoloaded just fine.
```

If the client library code sends the name of the class from global namespace with leading slash to `ComposerLocator`, locator will not be able to find this class!

PR #67, opened 2017-01-31
Fixed namespaces with starting slash

### LINKED ISSUE #42: Namespace \yii\base was not found in the file …/vendor/yiisoft/yii2/base/Application.php
I got error
  [Go\ParserReflection\ReflectionException]                                                             
  Namespace \yii\base was not found in the file …/vendor/yiisoft/yii2/base/Application.php                                                                                           

It parser has problem when parsing mine `\CommonApplication` (code below) that required parsing `\yii2\base\Application`.

```
<?php
abstract class CommonApplication extends yii\base\Application
{
}
```

adding `\ ` after `extends` like `abstract class CommonApplication extends \yii\base\Application` solves the problem

i have parser-reflection@1.1.0

Fedott:
Fixes #42

### DIFF
--- a/src/ReflectionEngine.php
+++ b/src/ReflectionEngine.php
@@ -229,6 +229,7 @@ class ReflectionEngine
      * @param string $namespaceName Namespace name
      *
      * @return Namespace_
+     * @throws ReflectionException
      */
     public static function parseFileNamespace($fileName, $namespaceName)
     {
@@ -239,7 +240,7 @@ class ReflectionEngine
                 continue;
             }
             $topLevelNodeName = $topLevelNode->name ? $topLevelNode->name->toString() : '';
-            if ($topLevelNodeName === $namespaceName || "\\{$topLevelNodeName}" === $namespaceName) {
+            if (ltrim($topLevelNodeName, '\\') === trim($namespaceName, '\\')) {
                 return $topLevelNode;
             }
         }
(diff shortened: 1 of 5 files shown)


aik099:
Please also add a test case.

src/ReflectionEngine.php
```diff
@@ -239,7 +239,7 @@ public static function parseFileNamespace($fileName, $namespaceName)
                 continue;
             }
             $topLevelNodeName = $topLevelNode->name ? $topLevelNode->name->toString() : '';
-            if ($topLevelNodeName === $namespaceName) {
+            if ($topLevelNodeName === $namespaceName || "\\{$topLevelNodeName}" === $namespaceName) {
```
lisachenko: It will be better to normalize this value before any additional actions in the method beginning. Eg.

```php
$namespaceName = ltrim($namespaceName, '\\');
```

Then we will reduce method complexity by avoiding extra check in the if expression.

aik099: Then we need to normalize both `$namespaceName` and `$topLevelNodeName` because both developer can supply `$namespaceName` with leading `\` and in analyzed PHP file namespace can be defined as `namespace \NamespaceExample;`.

Fedott:
Hi!
I added test for specific case from issue #44 and added normalize for both namespace on compare.

hijarian:
I suspect this PR fixes https://github.com/Codeception/AspectMock/issues/117 as well and supersedes my own PR https://github.com/goaop/parser-reflection/pull/66, would be great if you merge it.

lisachenko:
Hi! Thanks!

Issue #16, opened 2016-06-11
The "\Go\ParserReflection\ReflectionClass::getMethods" returns wrong instance objects


aik099:
The `\Go\ParserReflection\ReflectionClass::getMethods` method is used, then instead of returning all objects of `\Go\ParserReflection\ReflectionMethod` class it:
- returns that class methods using `\Go\ParserReflection\ReflectionMethod` class objects
- returns inherited methods using `\ReflectionMethod` class objects

![2016-06-11_2205](https://cloud.githubusercontent.com/assets/1277526/15986974/d9d18bea-3020-11e6-81a6-2b1ec49844c1.png)

This results in Fatal Error, when trying to process PHP 7 code using PHP 5.

Same for `getProperties` and maybe other methods that collect data from parent classes.

aik099:
The above described behavior happens due following code (2 places) where if class is autoloaded prior to library usage, then `\ReflectionClass` instead of `\Go\ParserReflection\ReflectionClass` is used:

* https://github.com/goaop/parser-reflection/blob/master/src/ReflectionParameter.php#L187-L189
* https://github.com/goaop/parser-reflection/blob/master/src/Traits/ReflectionClassLikeTrait.php#L448

This might be done like so to avoid running extra code, when we already got class autoloaded anyway. However such optimization results in inconsistent classes being returned out of the library.

I'm proposing to doing something like, class exists:

```php
$reflectionClass = new \ReflectionClass($className);
$reflectionFile = \Go\ParserReflection\ReflectionFile($reflectionClass->getFileName());

foreach ( $reflectionFile->getFileNamespaces() as $namespace ) {
    foreach ( $namespace->getClasses() as $class ) {
        if ( $class->getName() === $className ) {
            return $class;
        }
    }
}
```

lisachenko:
Hello, this library was designed in such a way to be as close to the original behavior as possible, so if class is already loaded into the memory (or internal one) then no need in parser reflection objects, because you will receive the same information from internal reflection much easier.

So, you shouldn't expect `ParserReflection` instances all the time. Of course, I can wrap all reflection objects with my own, but this will be simple proxying without any benefits and I decided to return just original objects instead.

aik099:
[...]

PR #69, opened 2017-02-04
Static ReflectionEngine replaced with injected ReflectionParser

dg:
Hi @lisachenko, thanks for great library!

I created a tool that adds missing PHP 7 typehints to codebase.  It uses Parser Reflection. This tools parses a lot of files (whole project) and checks whether all method's declarations are compatible with their prototype/interfaces. 

It is important that it uses **only** virtual AST-based class reflections and never fallback to existing classes (exception is native classes). ~~So I moved every class_exists() etc. checking to centrail point `ReflectionEngine::isClassEmulated()`.~~ So I replaced static `ReflectionEngine` with injected `ReflectionParser` and moved every class_exists() etc. checking to centrail point `ReflectionParser::getClassReflection()`

PR should be compatible with older behavior.

Now I can replace this rutine with my implementation:

```php
    public function getClassReflection($name)
    {
		if ((class_exists($name, false)
				|| interface_exists($name, false)
				|| trait_exists($name, false)
			) && !(new \ReflectionClass($name))->isUserDefined() // here is a difference
		) {
			return new \ReflectionClass($name);
		}
		return new ParserReflection\ReflectionClass($name, null, $this);
    }
```

~~It is not possible to inherit from ReflectionEngine, because all classes use static ReflectionEngine. One solution is to add something like `EmulatorDeciderInterface` with method `isEmulated` and inject it the same way as locator is injected.~~

~~Other idea is to leave the decision on locator itself. When it can locate class, use this one, when it can't use native one.~~ (Solved by injected `ReflectionParser`).

What do you think?

Idea: renamed ReflectionParser -> ReflectionContext

### DIFF
--- a/src/NodeVisitor/StaticVariablesCollector.php
+++ b/src/NodeVisitor/StaticVariablesCollector.php
@@ -10,6 +10,8 @@
 
 namespace Go\ParserReflection\NodeVisitor;
 
+use Go\ParserReflection\ReflectionEngine;
+use Go\ParserReflection\ReflectionParser;
 use Go\ParserReflection\ValueResolver\NodeExpressionResolver;
 use PhpParser\Node;
 use PhpParser\NodeTraverser;
@@ -27,6 +29,11 @@ class StaticVariablesCollector extends NodeVisitorAbstract
      */
     private $context;
 
+    /**
+     * @var ReflectionParser
+     */
+    private $reflectionParser;
+
     private $staticVariables = [];
 
     /**
@@ -34,9 +41,10 @@ class StaticVariablesCollector extends NodeVisitorAbstract
      *
      * @param mixed $context Reflection context, eg. ReflectionClass, ReflectionMethod, etc
      */
-    public function __construct($context)
+    public function __construct($context, ReflectionParser $reflectionParser = null)
     {
         $this->context = $context;
+        $this->reflectionParser = $reflectionParser ?: ReflectionEngine::getReflectionParser();
     }
 
     /**
@@ -50,7 +58,7 @@ class StaticVariablesCollector extends NodeVisitorAbstract
         }
 
         if ($node instanceof Node\Stmt\Static_) {
-            $expressionSolver = new NodeExpressionResolver($this->context);
+            $expressionSolver = new NodeExpressionResolver($this->context, $this->reflectionParser);
             $staticVariables  = $node->vars;
             foreach ($staticVariables as $staticVariable) {
                 $expr = $staticVariable->default;
--- a/src/ReflectionClass.php
+++ b/src/ReflectionClass.php
@@ -54,7 +54,7 @@ class ReflectionClass extends InternalReflectionClass
      *
      * @return array|\ReflectionClass[] List of reflections of interfaces
      */
-    public static function collectInterfacesFromClassNode(ClassLike $classLikeNode)
+    public static function collectInterfacesFromClassNode(ClassLike $classLikeNode, ReflectionParser $reflectionParser)
     {
         $interfaces = [];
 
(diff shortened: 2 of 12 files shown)


aik099:
> This tools parses a lot of files (whole project) and checks whether all method's declarations are compatible with their prototype/interfaces.

In other words you're checking code to avoid `E_STRICT` notice, when running the code?

In project I'm creating using this library (the static code analysis tool: https://github.com/console-helpers/code-insight) I'm also planning to create such feature: https://github.com/console-helpers/code-insight/issues/8. Can you please share your code that performs that check?

dg:
Related to #16

dg:
@aik099 Yes. I will release it in few days.

aik099:
Maybe you can post link to GitHub project so that I can watch to get notification about release?

dg:
Probably it will be named `updater` and located here https://github.com/nette

lisachenko:
Hi, @dg and @aik099! Thank you for the issue reporting and discussion!

I have several thoughts about current implementation: I'm trying to use AST information if possible, but if a class is already loaded, then in many places engine will return native implementation of `ReflectionClass` instead of AST-based to improve the performance. For me it's ok, but for some cases this can be improved, like in your example.

There is a request #2 to always return AST-based classes even for internal and loaded classes. This will require to add extra check in each reflection method `if ($this->isInitialized) { return parent::func(...$args); }`.

What if we put this logic into the static method `ReflectionEngine::isLoaded(string $fullClassName) : bool`?

Then in your code you can use it with the `ReflectionClass::isUserDefined()` method to perform extra logic.

dg:
For my use case is important, that even if class is loaded or not, it will return reflection that is based on parsed file, not in-memory-loaded class. 

So I can create tool that itself uses library ie. Nette\Utils in version 3.0.1 and is able to load in virtual storage project that uses the same named classes from Nette\Utils but in a different version.

Therefore word `isLoaded` sounds like something that will not help me.

dg:
The ideal solution is to replace static ReflectionEngine with object and pass it between other object, so you can create more than one "context" for parsed classes.

I know that there must exist fallback for standard usage like `$rc = new Go\ParserReflection\ReflectionClass($name)`, when default static context will be used.

lisachenko:
@dg, if you want to always use AST version, then you could use `ReflectionFile` to analyze any valid PHP file and extract a class from it, or even parse source code from the string and pass it as second argument to the `ReflectionFile`, it's also possible.

```php
$ast = ReflectionEngine::parseFile('virtual://any-file', '<?php echo "Yes";');
$file = new ReflectionFile('virtual://any-file', $ast);
```

Is it suitable for you? I use this technique inside `goaop/framework` to analyze a content from the string.

dg:
When I extract class from parsed file and call for example `getMethods()`, it will use "loaded" class parent instead of parent I want to use, right?

lisachenko:
It depends on information about loaded files and classes.

When you ask for `getMethods()`, engine checks all parents and asks for methods. If concrete class is already loaded, then native reflection will be used, otherwise a request to the locator will be performed to return a file name for AST analysis.

dg:
As I said, I never want to use loaded class, because it can be a different version of class than I want to use.

dg:
I'll try to prepare more complex PR.

dg:
@lisachenko I tried to create non-static replacement for static ReflectionEngine (it should be 100% compatible).

Because non-static class is injected where needed, I can replace it with own implementation and simply adjust Parser Reflection for my needs.

peter279k:
It looks like some files are conflict in this PR.

I think it should be fixed firstly.

lisachenko:
New year is a good time to clean all old issues and PR ) Closing this one

Issue #73, opened 2017-04-09
Locator needs to be aware of prerequisite classes/interfaces/traits


loren-osborn:
I figured I might as well report this as well, as I already have a fix, and am just awaiting permission to issue a PR.

`Go\ParserReflection\ReflectionClass` fails to find parent classes, implemented interfaces and used traits defined in the same file prior to the referencing class/interface/trait unless the locator is aware of which file to each prerequisite. While this isn't the recommended way to organize classes, etc. the ReflectionEngine should be able to handle it as it's relatively common, (especially for "packaged" modules) and it's perfectly valid PHP without an autoloader.

As I said above, I'm hoping to issue a PR for this soon.

loren-osborn:
This is a duplicate of issue #64, but I have a working fix for this that I just need permission to disclose and contribute.

Issue #72, opened 2017-04-07
implementsInterface() isSubclassOf() method inconsistancies in ReflectionClass


loren-osborn:
I'm finding differences in how `Go\ParserReflection\ReflectionClass` and native `ReflectionClass` behave for methods `implementsInterface()` and `isSubclassOf()`. I'm using some fakery to simulate external file loading, so I would be unsurprised if the issue involves how I'm populating these files, but I don't currently see anything I missing in that regard.

This is a couple pages of code, so it's hard to call it a "**minimal** reproducible test case", but I wrote it to be self-contained, and to demonstrate the issue I'm seeing in my code. I hope my comments are adequate, but I opted for clarity of the output over clarity of the code. This is why I chose to use ANSI color codes to make the issue easier to see.

As far as the issue I'm seeing in my actual codebase, I'm seeing the following behavior:

- `implementsInterface()` is working as expected.
- `isSubclassOf()` is always returning `FALSE`.

Here is a gist of the example code I'm using: https://gist.github.com/loren-osborn/b3a80c075ebb3ac6d3e2d8749a23bbe7 :

- Lines in green are working as expected.
- Lines in red (marked with `!!`) are returning `FALSE` when they should return `TRUE`.
- Lines with a yellow (marked with `..`) background are not throwing an exception that the native `ReflectionClass` method throws. This may or may not be the expected behavior. (I suggest that `Go\ParserReflection\ReflectionClass` should throw an exception as well, but that's my personal opinion.)

This is what my output looks like::
![goaop_parserreflection_issubclassofissue](https://cloud.githubusercontent.com/assets/2467372/24822576/ea4ea3ba-1baa-11e7-8f9d-7e7a3d0cf952.png)

loren-osborn:
I actually have developed a fix for this, but:

- ~~My earliest changes that resolved this issue didn't properly account for classes that had already been loaded and broke the existing unit tests.~~ My updated changes resolved this.
- I still need to convert this mini test suite into unit tests to add to PHPUnit
- I am awaiting permission before I can release any changes made on company time. (I think the odds of me getting permission soon are reasonably high.)

lisachenko:
Hi, @loren-osborn!

Thank you for reporting this issue. Of course, sometimes it can be very hard to emulate original reflection behavior, so I suppose that there are issues in my library. Some of them are important and should be fixed, whereas some of them aren't.

So, if you can provide broken PHPUnit test with PR then I'll be happy to review it and merge fix. 

[...]

Issue #64, opened 2016-12-30
Add support for analysing multiple classes in one file without asking the locator


onet4:
It seems calling `getMethods()` on a parsed child class causes `Fatal error: Uncaught exception 'InvalidArgumentException' with message 'xxx was not found by locator'`.

For example,

**Sample.php**
```php
<?php

namespace Foo\Bar;

abstract class TestBase {

    public function __construct() {
    }

}

class Test extends TestBase {

    public function __construct() {
        parent::__construct();
    }

}
```

**test.php**
```php
include( './vendor/autoload.php' );

$parsedFile     = new \Go\ParserReflection\ReflectionFile( __DIR__ . '/Sample.php');

foreach ($parsedFile->getFileNamespaces() as $namespace) {

    foreach ($namespace->getClasses() as $class) {
        foreach ($class->getMethods() as $method) {
            echo "Found class method: ", $class->getName(), '::', $method->getName(), PHP_EOL;
        }
    }

}
```

By running `test.php`, the line `        foreach ($class->getMethods() as $method) {` causes the above error.

lisachenko:
You are trying to analyze several classes, defined in one file. Currently it's not supported by the engine. You should use PSR standard and put each class into it's own file, so composer should know where to look for them.

Alternative way is to define your custom locator for that:

> If project uses a custom autoloader then you should follow the next steps:

>    Create a new class that implements \Go\ParserReflection\LocatorInterface
    Create an instance of that class and pass it to the ReflectionEngine::init() method for initial configuration

I want to convert this issue to the improvement request, but not for the next version. Engine will be able to look at parsed classes in the cache.

onet4:
Looking forward to it.

loren-osborn:
I'm thinking issue #73 is a duplicate of this, but I might mark it the other way around, as I already have a working solution for this. @lisachenko, what do you think? (I don't expect to have a PR publicly available for about 2 weeks... pending permission.)

Issue #17, opened 2016-06-12
The "\Go\ParserReflection\ReflectionEngine::locateClassFile" doesn't work with built-in classes


aik099:
When attempting to use `\Go\ParserReflection\ReflectionEngine::locateClassFile` on a built-in class (for them `\ReflectionClass::getFileName()` returns `false`), then exception about not found class is thrown.

I'm proposing to show `Locating of built-in classes is not supported.` exception message, when attempting to find built-in class.

lisachenko:
I can agree with this issue, but I have another vision: enable support for existing classes in all Reflection classes, so if class or function is already exists, then we can create a proxy for original reflection and forward all method calls to the internal reflection instances.

What do your think?

aik099:
Proxy sounds good.

aik099:
Not sure really how such proxy can't be implemented without overriding each method in each reflection class once again.

lisachenko:
Related to the #48 and #50

loren-osborn:
@lisachenko, I hope this isn't starting to sound like a tired refrain, but I do now have a working implementation of this "pass through mode" you described here. I'm sure if you had some code to look at, this would be a lot more welcome comment. While I am still waiting on permission to share my changes, I wanted to reassure you as to what I anticipate the chain of events will look like.

I've been working for a couple months on modifying AspectMock to allow effective unit testing of my client's Drupal 7 based custom modules. I am very close to having a first pass working solution. Neither my boss nor I are in a good situation to request permission to contribute my changes back upstream until I can demonstrate a cost benefit to the organization. Given my changes are reasonably extensive, I think the cost advantage of not supporting a custom fork are pretty clear; I really just need to demonstrate that my changes are an effective tool for testing our code, which management already sees the value in. I'm hoping to have management endorsed code to contribute back in the next 2-4 weeks. 

<hr />

That said, that brings me to why I'm writing today:

[...]

PR #76, opened 2017-07-23
Scrutinizer fixes

loren-osborn:
This PR addresses most of the issues highlighted by the Scrutinizer code analysis tool this package is already coupled to.

### DIFF
--- a/src/NodeVisitor/RootNamespaceNormalizer.php
+++ b/src/NodeVisitor/RootNamespaceNormalizer.php
@@ -23,7 +23,15 @@ use PhpParser\NodeVisitorAbstract;
 class RootNamespaceNormalizer extends NodeVisitorAbstract
 {
     /**
-     * {@inheritdoc}
+     * Called once before traversal.
+     *
+     * Return value semantics:
+     *  * null:      $nodes stays as-is
+     *  * otherwise: $nodes is set to the return value
+     *
+     * @param Node[] $nodes Array of nodes
+     *
+     * @return null|Node[] Array of nodes
      */
     public function beforeTraverse(array $nodes)
     {
--- a/src/NodeVisitor/RootNamespaceNormalizer.php
+++ b/src/NodeVisitor/RootNamespaceNormalizer.php
@@ -23,15 +23,7 @@ use PhpParser\NodeVisitorAbstract;
 class RootNamespaceNormalizer extends NodeVisitorAbstract
 {
     /**
-     * Called once before traversal.
-     *
-     * Return value semantics:
-     *  * null:      $nodes stays as-is
-     *  * otherwise: $nodes is set to the return value
-     *
-     * @param Node[] $nodes Array of nodes
-     *
-     * @return null|Node[] Array of nodes
+     * {@inheritdoc}
      */
     public function beforeTraverse(array $nodes)
     {
--- a/src/ReflectionFile.php
+++ b/src/ReflectionFile.php
@@ -51,7 +51,9 @@ class ReflectionFile
     public function __construct($fileName, $topLevelNodes = null)
     {
         if (!is_string($fileName)) {
-            throw new \InvalidArgumentException(sprintf('$fileName must be a string, but a %s was passed', gettype($fileName)));
+            throw new \InvalidArgumentException(sprintf(
+                '$fileName must be a string, but a %s was passed',
+                gettype($fileName)));
         }
         $fileName            = PathResolver::realpath($fileName);
         $this->fileName      = $fileName;
--- a/src/ReflectionFile.php
+++ b/src/ReflectionFile.php
@@ -50,6 +50,9 @@ class ReflectionFile
      */
     public function __construct($fileName, $topLevelNodes = null)
     {
+        if (!is_string($fileName)) {
+            throw new \InvalidArgumentException(sprintf('$fileName must be a string, but a %s was passed', gettype($fileName)));
+        }
         $fileName            = PathResolver::realpath($fileName);
         $this->fileName      = $fileName;
         $this->topLevelNodes = $topLevelNodes ?: ReflectionEngine::parseFile($fileName);
--- a/src/ReflectionFile.php
+++ b/src/ReflectionFile.php
@@ -51,9 +51,12 @@ class ReflectionFile
     public function __construct($fileName, $topLevelNodes = null)
     {
         if (!is_string($fileName)) {
-            throw new \InvalidArgumentException(sprintf(
-                '$fileName must be a string, but a %s was passed',
-                gettype($fileName)));
+            throw new \InvalidArgumentException(
+                sprintf(
+                    '$fileName must be a string, but a %s was passed',
+                    gettype($fileName)
+                )
+            );
         }
         $fileName            = PathResolver::realpath($fileName);
         $this->fileName      = $fileName;
--- a/src/ReflectionFileNamespace.php
+++ b/src/ReflectionFileNamespace.php
@@ -83,6 +83,9 @@ class ReflectionFileNamespace
      */
     public function __construct($fileName, $namespaceName, Namespace_ $namespaceNode = null)
     {
+        if (!is_string($fileName)) {
+            throw new \InvalidArgumentException(sprintf('$fileName must be a string, but a %s was passed', gettype($fileName)));
+        }
         $fileName = PathResolver::realpath($fileName);
         if (!$namespaceNode) {
             $namespaceNode = ReflectionEngine::parseFileNamespace($fileName, $namespaceName);
--- a/src/ReflectionFileNamespace.php
+++ b/src/ReflectionFileNamespace.php
@@ -84,9 +84,12 @@ class ReflectionFileNamespace
     public function __construct($fileName, $namespaceName, Namespace_ $namespaceNode = null)
     {
         if (!is_string($fileName)) {
-            throw new \InvalidArgumentException(sprintf(
-                '$fileName must be a string, but a %s was passed',
-                gettype($fileName)));
+            throw new \InvalidArgumentException(
+                sprintf(
+                    '$fileName must be a string, but a %s was passed',
+                    gettype($fileName)
+                )
+            );
         }
         $fileName = PathResolver::realpath($fileName);
         if (!$namespaceNode) {
--- a/src/ReflectionFileNamespace.php
+++ b/src/ReflectionFileNamespace.php
@@ -180,7 +180,7 @@ class ReflectionFileNamespace
         $comments   = $this->namespaceNode->getAttribute('comments');
 
         if ($comments) {
-            $docComment = (string) $comments[0];
+            $docComment = (string)$comments[0];
         }
 
         return $docComment;
--- a/src/ReflectionFileNamespace.php
+++ b/src/ReflectionFileNamespace.php
@@ -84,7 +84,9 @@ class ReflectionFileNamespace
     public function __construct($fileName, $namespaceName, Namespace_ $namespaceNode = null)
     {
         if (!is_string($fileName)) {
-            throw new \InvalidArgumentException(sprintf('$fileName must be a string, but a %s was passed', gettype($fileName)));
+            throw new \InvalidArgumentException(sprintf(
+                '$fileName must be a string, but a %s was passed',
+                gettype($fileName)));
         }
         $fileName = PathResolver::realpath($fileName);
         if (!$namespaceNode) {
--- a/src/ReflectionMethod.php
+++ b/src/ReflectionMethod.php
@@ -80,8 +80,8 @@ class ReflectionMethod extends BaseReflectionMethod
      */
     public function __toString()
     {
-        $hasReturnType    = $this->hasReturnType();
-        $returnType       = $hasReturnType ? $this->getReturnType() : null;
+        $returnType       = $this->getReturnType();
+        $hasReturnType    = ($returnType !== null); // Internally same as: $this->hasReturnType();
         $paramsNeeded     = $hasReturnType || $this->getNumberOfParameters() > 0;
         $paramFormat      = $paramsNeeded ? "\n\n  - Parameters [%d] {%s\n  }" : '';
         $returnFormat     = $hasReturnType ? "\n  - Return [ %s ]" : '';
@@ -108,14 +108,18 @@ class ReflectionMethod extends BaseReflectionMethod
             $this->isFinal() ? ' final' : '',
             $this->isStatic() ? ' static' : '',
             $this->isAbstract() ? ' abstract' : '',
-            join(' ', \Reflection::getModifierNames($this->getModifiers() & 1792)),
+            join(' ', \Reflection::getModifierNames(
+                $this->getModifiers() &
+                (self::IS_PUBLIC | self::IS_PROTECTED | self::IS_PRIVATE))),
             $this->getName(),
             $this->getFileName(),
             $this->getStartLine(),
             $this->getEndLine(),
             count($methodParameters),
             $paramString,
-            $returnType ? ReflectionType::convertToDisplayType($returnType) : ''
+            (   $returnType ?
+                ReflectionType::convertToDisplayType($returnType) :
+                'UNUSED: Prevents convertToDisplayType() being called.')
         );
     }
 
--- a/src/ReflectionMethod.php
+++ b/src/ReflectionMethod.php
@@ -81,6 +81,7 @@ class ReflectionMethod extends BaseReflectionMethod
     public function __toString()
     {
         $hasReturnType    = $this->hasReturnType();
+        $returnType       = $hasReturnType ? $this->getReturnType() : NULL;
         $paramsNeeded     = $hasReturnType || $this->getNumberOfParameters() > 0;
         $paramFormat      = $paramsNeeded ? "\n\n  - Parameters [%d] {%s\n  }" : '';
         $returnFormat     = $hasReturnType ? "\n  - Return [ %s ]" : '';
@@ -114,7 +115,7 @@ class ReflectionMethod extends BaseReflectionMethod
             $this->getEndLine(),
             count($methodParameters),
             $paramString,
-            $hasReturnType ? ReflectionType::convertToDisplayType($this->getReturnType()) : ''
+            $returnType ? ReflectionType::convertToDisplayType($returnType) : ''
         );
     }
 
--- a/src/ReflectionMethod.php
+++ b/src/ReflectionMethod.php
@@ -81,7 +81,7 @@ class ReflectionMethod extends BaseReflectionMethod
     public function __toString()
     {
         $hasReturnType    = $this->hasReturnType();
-        $returnType       = $hasReturnType ? $this->getReturnType() : NULL;
+        $returnType       = $hasReturnType ? $this->getReturnType() : null;
         $paramsNeeded     = $hasReturnType || $this->getNumberOfParameters() > 0;
         $paramFormat      = $paramsNeeded ? "\n\n  - Parameters [%d] {%s\n  }" : '';
         $returnFormat     = $hasReturnType ? "\n  - Return [ %s ]" : '';
--- a/src/ReflectionMethod.php
+++ b/src/ReflectionMethod.php
@@ -121,9 +121,7 @@ class ReflectionMethod extends BaseReflectionMethod
             $this->getEndLine(),
             count($methodParameters),
             $paramString,
-            (   $returnType ?
-                ReflectionType::convertToDisplayType($returnType) :
-                'UNUSED: Prevents convertToDisplayType() being called.')
+            ($returnType ? ReflectionType::convertToDisplayType($returnType) : '')
         );
     }
 
--- a/src/ReflectionMethod.php
+++ b/src/ReflectionMethod.php
@@ -80,8 +80,9 @@ class ReflectionMethod extends BaseReflectionMethod
      */
     public function __toString()
     {
+        // Internally $this->getReturnType() !== null is the same as $this->hasReturnType()
         $returnType       = $this->getReturnType();
-        $hasReturnType    = ($returnType !== null); // Internally same as: $this->hasReturnType();
+        $hasReturnType    = $returnType !== null;
         $paramsNeeded     = $hasReturnType || $this->getNumberOfParameters() > 0;
         $paramFormat      = $paramsNeeded ? "\n\n  - Parameters [%d] {%s\n  }" : '';
         $returnFormat     = $hasReturnType ? "\n  - Return [ %s ]" : '';
@@ -111,8 +112,7 @@ class ReflectionMethod extends BaseReflectionMethod
             join(
                 ' ',
                 \Reflection::getModifierNames(
-                    $this->getModifiers() &
-                        (self::IS_PUBLIC | self::IS_PROTECTED | self::IS_PRIVATE)
+                    $this->getModifiers() & (self::IS_PUBLIC | self::IS_PROTECTED | self::IS_PRIVATE)
                 )
             ),
             $this->getName(),
@@ -121,7 +121,7 @@ class ReflectionMethod extends BaseReflectionMethod
             $this->getEndLine(),
             count($methodParameters),
             $paramString,
-            ($returnType ? ReflectionType::convertToDisplayType($returnType) : '')
+            $returnType ? ReflectionType::convertToDisplayType($returnType) : ''
         );
     }
 
--- a/src/ReflectionMethod.php
+++ b/src/ReflectionMethod.php
@@ -108,9 +108,13 @@ class ReflectionMethod extends BaseReflectionMethod
             $this->isFinal() ? ' final' : '',
             $this->isStatic() ? ' static' : '',
             $this->isAbstract() ? ' abstract' : '',
-            join(' ', \Reflection::getModifierNames(
-                $this->getModifiers() &
-                (self::IS_PUBLIC | self::IS_PROTECTED | self::IS_PRIVATE))),
+            join(
+                ' ',
+                \Reflection::getModifierNames(
+                    $this->getModifiers() &
+                        (self::IS_PUBLIC | self::IS_PROTECTED | self::IS_PRIVATE)
+                )
+            ),
             $this->getName(),
             $this->getFileName(),
             $this->getStartLine(),
--- a/src/ReflectionType.php
+++ b/src/ReflectionType.php
@@ -85,7 +85,7 @@ class ReflectionType extends BaseReflectionType
             'int'  => 'integer',
             'bool' => 'boolean'
         ];
-        $displayType = (string)$type;
+        $displayType = (string) $type;
         if (isset($typeMap[$displayType])) {
             $displayType = $typeMap[$displayType];
         }
--- a/src/ReflectionType.php
+++ b/src/ReflectionType.php
@@ -92,7 +92,7 @@ class ReflectionType extends BaseReflectionType
 
         $displayType = ltrim($displayType, '\\');
 
-        if (method_exists($type, 'allowsNull') && $type->allowsNull()) {
+        if ($type->allowsNull()) {
             $displayType .= ' or NULL';
         }
 
--- a/src/ReflectionType.php
+++ b/src/ReflectionType.php
@@ -85,7 +85,7 @@ class ReflectionType extends BaseReflectionType
             'int'  => 'integer',
             'bool' => 'boolean'
         ];
-        $displayType = (string) $type;
+        $displayType = (string)$type;
         if (isset($typeMap[$displayType])) {
             $displayType = $typeMap[$displayType];
         }
--- a/src/ReflectionType.php
+++ b/src/ReflectionType.php
@@ -85,14 +85,14 @@ class ReflectionType extends BaseReflectionType
             'int'  => 'integer',
             'bool' => 'boolean'
         ];
-        $displayType = $type->type;
+        $displayType = (string)$type;
         if (isset($typeMap[$displayType])) {
             $displayType = $typeMap[$displayType];
         }
 
         $displayType = ltrim($displayType, '\\');
 
-        if ($type->allowsNull()) {
+        if (method_exists($type, 'allowsNull') && $type->allowsNull()) {
             $displayType .= ' or NULL';
         }
 
--- a/src/Traits/ReflectionClassLikeTrait.php
+++ b/src/Traits/ReflectionClassLikeTrait.php
@@ -444,7 +444,6 @@ trait ReflectionClassLikeTrait
             if ($extendsNode instanceof FullyQualified) {
                 $extendsName = $extendsNode->toString();
                 $parentClass = class_exists($extendsName, false) ? new BaseReflectionClass($extendsName) : new ReflectionClass($extendsName);
-
             }
             $this->parentClass = $parentClass;
         }
--- a/src/Traits/ReflectionClassLikeTrait.php
+++ b/src/Traits/ReflectionClassLikeTrait.php
@@ -11,6 +11,7 @@
 namespace Go\ParserReflection\Traits;
 
 use Go\ParserReflection\ReflectionClass;
+use ReflectionClass as BaseReflectionClass;
 use Go\ParserReflection\ReflectionException;
 use Go\ParserReflection\ReflectionMethod;
 use Go\ParserReflection\ReflectionProperty;
@@ -442,7 +443,8 @@ trait ReflectionClassLikeTrait
             $extendsNode = $hasExtends ? $this->classLikeNode->$extendsField : null;
             if ($extendsNode instanceof FullyQualified) {
                 $extendsName = $extendsNode->toString();
-                $parentClass = class_exists($extendsName, false) ? new parent($extendsName) : new static($extendsName);
+                $parentClass = class_exists($extendsName, false) ? new BaseReflectionClass($extendsName) : new ReflectionClass($extendsName);
+
             }
             $this->parentClass = $parentClass;
         }
@@ -698,8 +700,16 @@ trait ReflectionClassLikeTrait
         }
 
         $className = $this->getName();
-
-        return $className === get_class($object) || is_subclass_of($object, $className);
+        $objClass = get_class($object);
+        if ($className === $objClass) {
+            return true;
+        }
+        $objClassRef = new BaseReflectionClass($objClass);
+        if ($objClassRef->isSubclassOf($className)) {
+            return true;
+        }
+        $thisIsInterface = $this->isInterface();
+        return $thisIsInterface && $objClassRef->implementsInterface($className);
     }
 
     /**
--- a/src/Traits/ReflectionClassLikeTrait.php
+++ b/src/Traits/ReflectionClassLikeTrait.php
@@ -443,7 +443,7 @@ trait ReflectionClassLikeTrait
             $extendsNode = $hasExtends ? $this->classLikeNode->$extendsField : null;
             if ($extendsNode instanceof FullyQualified) {
                 $extendsName = $extendsNode->toString();
-                $parentClass = class_exists($extendsName, false) ? new BaseReflectionClass($extendsName) : new ReflectionClass($extendsName);
+                $parentClass = $this->createReflectionForClass($extendsName);
             }
             $this->parentClass = $parentClass;
         }
@@ -699,16 +699,7 @@ trait ReflectionClassLikeTrait
         }
 
         $className = $this->getName();
-        $objClass = get_class($object);
-        if ($className === $objClass) {
-            return true;
-        }
-        $objClassRef = new BaseReflectionClass($objClass);
-        if ($objClassRef->isSubclassOf($className)) {
-            return true;
-        }
-        $thisIsInterface = $this->isInterface();
-        return $thisIsInterface && $objClassRef->implementsInterface($className);
+        return $className === get_class($object) || is_subclass_of($object, $className);
     }
 
     /**
@@ -963,4 +954,6 @@ trait ReflectionClassLikeTrait
             }
         }
     }
+
+    abstract protected function createReflectionForClass($className);
 }
--- a/src/Traits/ReflectionClassLikeTrait.php
+++ b/src/Traits/ReflectionClassLikeTrait.php
@@ -11,7 +11,6 @@
 namespace Go\ParserReflection\Traits;
 
 use Go\ParserReflection\ReflectionClass;
-use ReflectionClass as BaseReflectionClass;
 use Go\ParserReflection\ReflectionException;
 use Go\ParserReflection\ReflectionMethod;
 use Go\ParserReflection\ReflectionProperty;
@@ -323,6 +322,7 @@ trait ReflectionClassLikeTrait
 
     /**
      * {@inheritdoc}
+     * @param string $name
      */
     public function getMethod($name)
     {
@@ -593,6 +593,7 @@ trait ReflectionClassLikeTrait
 
     /**
      * {@inheritdoc}
+     * @param string $name
      */
     public function hasMethod($name)
     {
@@ -623,6 +624,7 @@ trait ReflectionClassLikeTrait
 
     /**
      * {@inheritDoc}
+     * @param string $interfaceName
      */
     public function implementsInterface($interfaceName)
     {
@@ -955,5 +957,14 @@ trait ReflectionClassLikeTrait
         }
     }
 
+    /**
+     * Create a ReflectionClass for a given class name.
+     *
+     * @param string $className
+     *     The name of the class to create a reflection for.
+     *
+     * @return ReflectionClass
+     *     The apropriate reflection object.
+     */
     abstract protected function createReflectionForClass($className);
 }
--- a/src/ValueResolver/NodeExpressionResolver.php
+++ b/src/ValueResolver/NodeExpressionResolver.php
@@ -258,8 +258,7 @@ class NodeExpressionResolver
         $classToReflect = $node->class;
         if ($classToReflect instanceof Expr) {
             $refClass = $this->resolve($classToReflect);
-        }
-        else {
+        } else {
             $refClass = $this->fetchReflectionClass($node->class);
         }
         $constantName = $node->name;
--- a/src/ValueResolver/NodeExpressionResolver.php
+++ b/src/ValueResolver/NodeExpressionResolver.php
@@ -112,8 +112,7 @@ class NodeExpressionResolver
         try {
             ++$this->nodeLevel;
 
-            $nodeType   = $node->getType();
-            $methodName = 'resolve' . str_replace('_', '', $nodeType);
+            $methodName = $this->getDispatchMethodFor($node);
             if (method_exists($this, $methodName)) {
                 $value = $this->$methodName($node);
             }
@@ -142,7 +141,7 @@ class NodeExpressionResolver
     protected function resolveScalarMagicConstMethod()
     {
         if ($this->context instanceof \ReflectionMethod) {
-            $fullName = $this->context->class . '::' . $this->context->getShortName();
+            $fullName = $this->context->getDeclaringClass()->name . '::' . $this->context->getShortName();
 
             return $fullName;
         }
@@ -231,7 +230,6 @@ class NodeExpressionResolver
             if (method_exists($this->context, 'getFileName')) {
                 $fileName      = $this->context->getFileName();
                 $namespaceName = $this->resolveScalarMagicConstNamespace();
-                /** @var ReflectionFileNamespace $fileNamespace */
                 $fileNamespace = new ReflectionFileNamespace($fileName, $namespaceName);
                 if ($fileNamespace->hasConstant($constantName)) {
                     $constantValue = $fileNamespace->getConstant($constantName);
@@ -256,11 +254,21 @@ class NodeExpressionResolver
     protected function resolveExprClassConstFetch(Expr\ClassConstFetch $node)
     {
         $classToReflect = $node->class;
-        if ($classToReflect instanceof Expr) {
-            $refClass = $this->resolve($classToReflect);
-        } else {
-            $refClass = $this->fetchReflectionClass($node->class);
+        if (!($classToReflect instanceof Node\Name)) {
+            $classToReflect = $this->resolve($classToReflect) ?: $classToReflect;
+            if (!is_string($classToReflect)) {
+                $reason = 'Unable';
+                if ($classToReflect instanceof Expr) {
+                    $methodName = $this->getDispatchMethodFor($classToReflect);
+                    $reason = "Method " . __CLASS__ . "::{$methodName}() not found trying";
+                }
+                throw new ReflectionException("$reason to resolve class constant.");
+            }
+            // Strings evaluated as class names are always treated as fully
+            // qualified.
+            $classToReflect = new Node\Name\FullyQualified(ltrim($classToReflect, '\\'));
         }
+        $refClass = $this->fetchReflectionClass($classToReflect);
         $constantName = $node->name;
 
         // special handling of ::class constants
@@ -269,7 +277,7 @@ class NodeExpressionResolver
         }
 
         $this->isConstant = true;
-        $this->constantName = (string)$node->class . '::' . $constantName;
+        $this->constantName = (string)$classToReflect . '::' . $constantName;
 
         return $refClass->getConstant($constantName);
     }
@@ -432,6 +440,12 @@ class NodeExpressionResolver
     protected function resolveExprBinaryOpLogicalXor(Expr\BinaryOp\LogicalXor $node)
     {
         return $this->resolve($node->left) xor $this->resolve($node->right);
+    }
+
+    private function getDispatchMethodFor(Node $node)
+    {
+        $nodeType = $node->getType();
+        return 'resolve' . str_replace('_', '', $nodeType);
     }
 
     /**
--- a/src/ValueResolver/NodeExpressionResolver.php
+++ b/src/ValueResolver/NodeExpressionResolver.php
@@ -142,7 +142,7 @@ class NodeExpressionResolver
     protected function resolveScalarMagicConstMethod()
     {
         if ($this->context instanceof \ReflectionMethod) {
-            $fullName = $this->context->getDeclaringClass()->getName() . '::' . $this->context->getShortName();
+            $fullName = $this->context->class . '::' . $this->context->getShortName();
 
             return $fullName;
         }
@@ -175,12 +175,12 @@ class NodeExpressionResolver
     protected function resolveScalarMagicConstClass()
     {
         if ($this->context instanceof \ReflectionClass) {
-            return $this->context->getName();
+            return $this->context->name;
         }
         if (method_exists($this->context, 'getDeclaringClass')) {
             $declaringClass = $this->context->getDeclaringClass();
             if ($declaringClass instanceof \ReflectionClass) {
-                return $declaringClass->getName();
+                return $declaringClass->name;
             }
         }
 
@@ -213,7 +213,7 @@ class NodeExpressionResolver
     protected function resolveScalarMagicConstTrait()
     {
         if ($this->context instanceof \ReflectionClass && $this->context->isTrait()) {
-            return $this->context->getName();
+            return $this->context->name;
         }
 
         return '';
@@ -224,8 +224,6 @@ class NodeExpressionResolver
         $constantValue = null;
         $isResolved    = false;
 
-        /** @var ReflectionFileNamespace|null $fileNamespace */
-        $fileNamespace = null;
         $isFQNConstant = $node->name instanceof Node\Name\FullyQualified;
         $constantName  = $node->name->toString();
 
@@ -233,6 +231,7 @@ class NodeExpressionResolver
             if (method_exists($this->context, 'getFileName')) {
                 $fileName      = $this->context->getFileName();
                 $namespaceName = $this->resolveScalarMagicConstNamespace();
+                /** @var ReflectionFileNamespace $fileNamespace */
                 $fileNamespace = new ReflectionFileNamespace($fileName, $namespaceName);
                 if ($fileNamespace->hasConstant($constantName)) {
                     $constantValue = $fileNamespace->getConstant($constantName);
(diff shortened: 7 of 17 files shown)


aik099:
Isn't that too much? I mean changes made are protection code added for theoretical misuse cases of the library that never happen in reality. Placing protection code only suggests that bad thing are expected to happen.

src/polyfill.php
```diff
@@ -0,0 +1,23 @@
...
+ * This source file is subject to the license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+/**
+ * This file is for ployfilling classes not defined in all supported
+ * versions of PHP, (i.e. PHP < 7).
+ */
+if (!class_exists(ReflectionType::class, false)) {
+    class ReflectionType
```
loren-osborn: This is obviously a non-issue for a polyfill class.

loren-osborn:
@aik099, My intent was really just good code hygiene. I certainly didn't want to impact performance. For several of the string argument checking, perhaps a docblock annotation is a better resolution?

loren-osborn:
Ok... **Update3**: I'm working on improving test coverage, but it's a bit more drastic than I hoped, so I have decided to put it in another PR that I will submit later. 

 @lisachenko , @aik099 can you give this a look, and let me know what you think, as it stands right now? @aik099 pointed out that there is some code that somewhat dubiously bullet proofs agains (hopefully) impossible cases, but the goal was to increase the signal-to-noise ratio in the Scrutinizer report, so it's easier to spot **real** issues when they arise. Additionally, someone with project write-access should also disable the issues that were intentional violations. (For instance, the `ReflectionType` polyfill class was **intentionally** put in the Global namespace.)  **If** any of the changes I proposed causes a performance issue, that **is** a _real concern I want to address_.

None of this is anything earth-shattering. It is intended to help keep the project clean and contributor-friendly. I welcome hearing any reason these changes should **not** be merged in. My intent was merely to help out.

**An aside**: I mentioned several improvements I've made to this project (and Codeception/AspectMock) for a former employer that I wanted to share. I am no longer with this company, and they have declined any interest in contributing these in the foreseeable future, despite my urging. They did mention they **may** be willing to release these for a fee. No amount was discussed, but I suggested a hypothetical future employer of mine might want to pay "a few grand" (that's a few $1000 USD to non-native speakers) for such changes, that could save them several months of my salary to reproduce such features. I have no idea how receptive to this idea this company still is, but if any of you want to raise some funds for this, I can help you get in touch with those in a position to discuss this.

**Anyway**, I hope you find this PR helpful.

aik099:
> An aside: I mentioned several improvements I've made to this project (and Codeception/AspectMock) for a former employer that I wanted to share. I am no longer with this company, and they have declined any interest in contributing these in the foreseeable future, despite my urging. They did mention they may be willing to release these for a fee. No amount was discussed, but I suggested a hypothetical future employer of mine might want to pay "a few grand" (that's a few $1000 USD to non-native speakers) for such changes, that could save them several months of my salary to reproduce such features. I have no idea how receptive to this idea this company still is, but if any of you want to raise some funds for this, I can help you get in touch with those in a position to discuss releasing those changes.

Can you list the changes included in there, so that we'll know what's coming.

aik099:
> @lisachenko , @aik099 can you give this a look, and let me know what you think, as it stands right now? @aik099 pointed out that there is some code that somewhat dubiously bullet proofs agains (hopefully) impossible cases, but the goal was to increase the signal-to-noise ratio in the Scrutinizer report, so it's easier to spot real issues when they arise. Additionally, someone with project write-access should also disable the issues that were intentional violations. (For instance, the ReflectionType polyfill class was intentionally put in the Global namespace.) If any of the changes I proposed causes a performance issue, that is a real concern I want to address.

Getting higher score on Scrutinizer CI won't be something worth reaching 😉  It's known for some weird rating calculating logic where some good classes are getting small ratings and attempt to fix rating results in making classes worse.

In my projects I'm reviewing each of the errors and decide if that is even worth fixing or I can simply ignore that. Not sure if ignoring errors would improve my rating though. 😄

aik099:
By the way I wonder why Scrutinizer CI isn't posting it's build status for this PR. Maybe this feature needs to be enabled manually.

src/ReflectionType.php
```diff
@@ -85,14 +85,14 @@ public static function convertToDisplayType(\ReflectionType $type)
             'int'  => 'integer',
             'bool' => 'boolean'
         ];
-        $displayType = $type->type;
+        $displayType = (string)$type;
```
aik099: 👍 , but a test is needed (or post link to existing test), that would make a call to `convertToDisplayType` method with:

1. native `ReflectionType` class object
2. with library's own `ReflectionType` object, that has `type` property

src/ReflectionType.php
```diff
@@ -85,14 +85,14 @@ public static function convertToDisplayType(\ReflectionType $type)
...
         if (isset($typeMap[$displayType])) {
             $displayType = $typeMap[$displayType];
         }
 
         $displayType = ltrim($displayType, '\\');
 
-        if ($type->allowsNull()) {
+        if (method_exists($type, 'allowsNull') && $type->allowsNull()) {
```
aik099: Please return the original code, because Scrutinizer CI issue in https://scrutinizer-ci.com/g/goaop/parser-reflection/inspections/eaf283f6-f982-44a4-bb28-51489bcba537/issues/files/src/ReflectionType.php?status=fixed&orderField=path&order=asc&honorSelectedPaths=0 likely is reported because they're using incorrect PHP stubs for class method signature resolution. On the PHP docs (see http://php.net/manual/en/reflectiontype.allowsnull.php) there is `allowsNull` method and it always was there (as in out `ReflectionType` class polyfill).

src/Traits/ReflectionClassLikeTrait.php
```diff
@@ -442,7 +443,7 @@ public function getParentClass()
             $extendsNode = $hasExtends ? $this->classLikeNode->$extendsField : null;
             if ($extendsNode instanceof FullyQualified) {
                 $extendsName = $extendsNode->toString();
-                $parentClass = class_exists($extendsName, false) ? new parent($extendsName) : new static($extendsName);
+                $parentClass = class_exists($extendsName, false) ? new BaseReflectionClass($extendsName) : new ReflectionClass($extendsName);
```
aik099: Please revert, because Scrutinizer CI seem to incorrectly tracked trait usage in https://scrutinizer-ci.com/g/goaop/parser-reflection/inspections/eaf283f6-f982-44a4-bb28-51489bcba537/issues/files/src/Traits/ReflectionClassLikeTrait.php?status=fixed&orderField=path&order=asc&honorSelectedPaths=0.

The original code doesn't have any issues.

src/Traits/ReflectionClassLikeTrait.php
```diff
@@ -698,8 +699,16 @@ public function isInstance($object)
         }
 
         $className = $this->getName();
-
-        return $className === get_class($object) || is_subclass_of($object, $className);
+        $objClass = get_class($object);
```
aik099: Please revert because Scrutinizer CI report (see https://scrutinizer-ci.com/g/goaop/parser-reflection/inspections/eaf283f6-f982-44a4-bb28-51489bcba537/issues/files/src/Traits/ReflectionClassLikeTrait.php?status=fixed&orderField=path&order=asc&honorSelectedPaths=0) mentions https://bugs.php.net/bug.php?id=53727 bug, which was fixed in PHP 5.3 and PHP 5.4, but this library has PHP 5.5+ as minimum required version.

Therefore bug never manifested itself in here.

src/ValueResolver/NodeExpressionResolver.php
```diff
@@ -142,7 +142,7 @@ protected function resolveScalarString(Scalar\String_ $node)
     protected function resolveScalarMagicConstMethod()
     {
         if ($this->context instanceof \ReflectionMethod) {
-            $fullName = $this->context->getDeclaringClass()->getName() . '::' . $this->context->getShortName();
+            $fullName = $this->context->class . '::' . $this->context->getShortName();
```
aik099: The suggestion in Scrutinizer CI inspection (see https://scrutinizer-ci.com/g/goaop/parser-reflection/inspections/eaf283f6-f982-44a4-bb28-51489bcba537/issues/files/src/ValueResolver/NodeExpressionResolver.php?status=fixed&orderField=path&order=asc&honorSelectedPaths=0) seems to be incorrect because mentioned https://bugs.php.net/bug.php?id=61384 PHP bug only affects `getName` method, but not `getDeclaringClass` method.

Knowing that I recommend

* replacing `$this->context->getDeclaringClass()->getName()`
* into `$this->context->getDeclaringClass()->name` instead.

src/ValueResolver/NodeExpressionResolver.php
```diff
@@ -175,12 +175,12 @@ protected function resolveScalarMagicConstNamespace()
     protected function resolveScalarMagicConstClass()
     {
         if ($this->context instanceof \ReflectionClass) {
-            return $this->context->getName();
+            return $this->context->name;
```
aik099: Looks good.

src/ValueResolver/NodeExpressionResolver.php
```diff
@@ -175,12 +175,12 @@ protected function resolveScalarMagicConstNamespace()
...
         }
         if (method_exists($this->context, 'getDeclaringClass')) {
             $declaringClass = $this->context->getDeclaringClass();
             if ($declaringClass instanceof \ReflectionClass) {
-                return $declaringClass->getName();
+                return $declaringClass->name;
```
aik099: Looks good.

src/ValueResolver/NodeExpressionResolver.php
```diff
@@ -213,7 +213,7 @@ protected function resolveScalarMagicConstLine(MagicConst\Line $node)
     protected function resolveScalarMagicConstTrait()
     {
         if ($this->context instanceof \ReflectionClass && $this->context->isTrait()) {
-            return $this->context->getName();
+            return $this->context->name;
```
aik099: Looks good.

src/ValueResolver/NodeExpressionResolver.php
```diff
@@ -224,15 +224,14 @@ protected function resolveExprConstFetch(Expr\ConstFetch $node)
...
-        /** @var ReflectionFileNamespace|null $fileNamespace */
-        $fileNamespace = null;
         $isFQNConstant = $node->name instanceof Node\Name\FullyQualified;
         $constantName  = $node->name->toString();
 
         if (!$isFQNConstant) {
             if (method_exists($this->context, 'getFileName')) {
                 $fileName      = $this->context->getFileName();
                 $namespaceName = $this->resolveScalarMagicConstNamespace();
+                /** @var ReflectionFileNamespace $fileNamespace */
```
aik099: You can safely remove this, because IDE can detect type from `new ...` construct by itself :wink:

src/ValueResolver/NodeExpressionResolver.php
```diff
@@ -256,7 +255,12 @@ protected function resolveExprConstFetch(Expr\ConstFetch $node)
 
     protected function resolveExprClassConstFetch(Expr\ClassConstFetch $node)
     {
-        $refClass     = $this->fetchReflectionClass($node->class);
+        $classToReflect = $node->class;
```
aik099: Doesn't seem to be related to Scrutinizer CI reported problems. Maybe belongs to separate PR with a description.

aik099:
@loren-osborn , review completed with outcome: don't trust Scrutinizer CI proposed fixes blindly 😉

src/ValueResolver/NodeExpressionResolver.php
loren-osborn: I'm  not sure why this issue isn't showing up for you, nor why I can't find links to individual Scrutinizer issues, but this is to address the Scrutinizer issue reported for this line that reads:

> <div><span style="color=white;background-color=grey">Bug</span> introduced 11 months ago by @lisachenko </div>
> It seems like <code style="color:red;">$node->class</code> can also be of type <code style="color:red;">object&lt;PhpParser\Node\Expr&gt;</code>; however, <code style="color:red;">Go\ParserReflection\Valu<span style="color=blue">...</span>:fetchReflectionClass()</code> does only seem to accept <code style="color:red;">object&lt;PhpParser\Node\Name&gt;</code>, maybe add an additional type check?

aik099: Interesting. So class constant can be an expression. Seem to addition in PHP syntax, that PHP-Parser now supports. In either cases a test for that is needed too.

src/NodeVisitor/RootNamespaceNormalizer.php
```diff
@@ -23,7 +23,15 @@
...
-     * {@inheritdoc}
+     * Called once before traversal.
+     *
+     * Return value semantics:
+     *  * null:      $nodes stays as-is
+     *  * otherwise: $nodes is set to the return value
+     *
+     * @param Node[] $nodes Array of nodes
+     *
+     * @return null|Node[] Array of nodes
```
loren-osborn: This is really a failed attempt to fix a false-positive in Scrutinizer. I merely copied the original function docblock from PhpParser. The issue is that Scrutinizer is letting the less specific type hint `array` take precedence over the compatible and more specific `@param Node[]`, and saying that `array` does not satisfy the specification `@return null|Node[]`. I expect this change should actually be reverted, but I didn't want to do so without feedback.

As far as the false-positive, I'm trying to notify Scrutinizer about it, but haven't gotten any confirmation yet.

src/ValueResolver/NodeExpressionResolver.php
loren-osborn: Yes, I know you can do a constant fetch from a class name stored in a variable. I'm unsure what other `Expr`s are valid. I believe *ParserReflection* uses this mostly for evaluating constants, so this might not be a valid code path as we are using it, but it is a valid value for `$node->class` in some circumstances.

As far as adding corresponding test cases, I agree, but I was lumping my test changes into a separate PR. (I'm shooting for 100% code coverage, but I've always been a cockeyed optimist.) I can add these tests to this PR after I've untangled my current repo branches. (I don't really want to do a stash right now.)

src/ValueResolver/NodeExpressionResolver.php
loren-osborn: Agreed... This comment just "came along for the ride" when I moved the variable's initialization.

src/NodeVisitor/RootNamespaceNormalizer.php
loren-osborn: Thumbing myself up purely so I remember to revert this change: 👍

loren-osborn:
@aik099, I'm sorry that I'm attacking this code review a bit piecemeal. I will push a new commit when I'm finished... In the meantime:

> By the way I wonder why Scrutinizer CI isn't posting it's build status for this PR. Maybe this feature needs to be enabled manually.

I actually posted a link to this above. Please let me know if this isn't valid: [Scrutinizer-CI issue report for PR #76 ](https://scrutinizer-ci.com/g/goaop/parser-reflection/inspections/eaf283f6-f982-44a4-bb28-51489bcba537/issues/)

src/ReflectionMethod.php
```diff
@@ -81,6 +81,7 @@ public function ___debugInfo()
     public function __toString()
     {
         $hasReturnType    = $this->hasReturnType();
+        $returnType       = $hasReturnType ? $this->getReturnType() : null;
```
loren-osborn: Good catch 👍

src/ReflectionMethod.php
```diff
@@ -114,7 +115,7 @@ public function __toString()
             $this->getEndLine(),
             count($methodParameters),
             $paramString,
-            $hasReturnType ? ReflectionType::convertToDisplayType($this->getReturnType()) : ''
+            $returnType ? ReflectionType::convertToDisplayType($returnType) : ''
```
loren-osborn: Ok... Good point on adding tests. As I said, needing to untangle a branch first, but: Note to self... Implement/find relevant test cases. 👍

src/ReflectionType.php
loren-osborn: As above... add tests 👍

src/ReflectionType.php
loren-osborn: Hmm... I agree that I should revert this code, but for a different reason. I think this is catching the class definition from the polyfill, which had no `$this-&gt;allowsNull()` method. I actually **added** a dummy `$this-&gt;allowsNull()` method to my polly fill, so this should no longer be an issue. I don't particularly like the dummy method in the polyfill, but I think it's the best solution currently. Either way, this code should no longer be necessary. Thanks for pointing this out. 👍

src/Traits/ReflectionClassLikeTrait.php
loren-osborn: Hmm... Here I'm a bit torn here:

- I don't like the old code: Scrutinize is right.
  + This is a trait. in theory, it should be includable in any class
  + Traits don't have parents, or even self.
    * PHP did the right thing in interpreting `self` and `parent` relative to the trait user
  + This interpretation of `self` and `parent` is rather non-obvious, and does nothing to aid code clarity.
  + There is no way to statically analyze this without the class it's being used in, which explains Scrutinize's issue.
- I don't like the new code either:
  + My changes totally lock this down to inclusion in one specific class: That's wrong too.
  + There really should be specified by the including class. a variable could work, but I think a class method in the using class is the best solution.
- One of the PRs I'm hoping to submit soon is a factory for reflection classes:
  + ... so I expect this issue to go away soon.

@aik099, for now, I think the best solution is a method in the using class to return the newly created objects. What do you think?

src/Traits/ReflectionClassLikeTrait.php
loren-osborn: This makes sense... I should check if there's a way to turn off a specific issue check for a whole project.

src/ValueResolver/NodeExpressionResolver.php
aik099: > Yes, I know you can do a constant fetch from a class name stored in a variable. I'm unsure what other Exprs are valid. 

I've meant `Example #4 Constant expression example` from http://php.net/manual/en/language.oop5.constants.php page:

```php
const ONE = 1;

class foo {
    // As of PHP 5.6.0
    const TWO = ONE * 2;
    const THREE = ONE + self::TWO;
    const SENTENCE = 'The value of THREE is '.self::THREE;
}
```

Working since PHP 5.6 only. This way you can create test for that kind of syntax.

> As far as adding corresponding test cases, I agree, but I was lumping my test changes into a separate PR. 

The tests for changed code needs to come as part of PR, where that code is changed. Later you can send another PR for improving code coverage.

aik099:
> I actually posted a link to this above. Please let me know if this isn't valid: Scrutinizer-CI issue report for PR #76

Perfectly fine. I was thinking about proper integration with GitHub where PR build status would fail if issues are found by Scrutinizer CI.

src/ReflectionType.php
aik099: > I think this is catching the class definition from the polyfill, which had no $this->allowsNull() method.

Then I think you need to add stubs in the polyfill class for all 3 methods: http://php.net/manual/en/class.reflectiontype.php

src/Traits/ReflectionClassLikeTrait.php
aik099: Fact, that trait is used only once and seems like it would be for a long time (unless PHP decides to reflect class/interface/trait using different reflection classes) is a problem that I think needs to be solved.

If we inline trait (copy trait code into the only class using it and drop trait), then problem should go away by itself. This is a major change however (even though not a BC break) and likely a @lisachenko approval is required before making such change.

Actually better to move it out into PR of it's own.

src/Traits/ReflectionClassLikeTrait.php
aik099: Users with administrative permissions for repo can do it on Scrutinizer CI.

src/ReflectionType.php
loren-osborn: Fair enough! :-) :+1:

lisachenko:
> Perfectly fine. I was thinking about proper integration with GitHub where PR build status would fail if issues are found by Scrutinizer CI.

Hi!
Just enabled integration of Scrutinizer-ci for PRs (previously it was enabled for commit statuses)

lisachenko:
So, guys, is this ready for merge?

src/Traits/ReflectionClassLikeTrait.php
loren-osborn: I was only really suggesting adding a  `protected function createReflectionObjectForClass()` containing only that one trinary expression line. Seems like a pretty minor change to me.

loren-osborn:
@lisachenko, thanks, but not ready to merge yet... I've only gone through two-thirds of @aik099's code review. I still need to make suggested changes to PR. 

When the PR **is** would you like these squashed into a single commit?

lisachenko:
> When the PR is would you like these squashed into a single commit?

This (squash) will be done automatically during the merge, no need to squash it before.

aik099:
> So, guys, is this ready for merge?

Your response is needed about inlining `ReflectionClassLikeTrait.php` (see comment about that, I don't know how to link into code comment directly).

loren-osborn:
> Your response is needed about inlining `ReflectionClassLikeTrait.php` (see comment about that, I don't know how to link into code comment directly).

@aik099, @lisachenko, I'm not opposed to inlining the whole trait, if that's the direction we deem appropriate, but I was only suggesting adding a single protected method to `Go\ParserReflection\ReflectionClass` named something like `createReflectionClassFor($className)`.

aik099:
@loren-osborn , you seem to have closed this PR by mistake. This usually happens when you rename a branch and push changes.

lisachenko:
Gyus! ) Stop blaming PR and let's move to the merge part, it's impossible to do everything just in one single PR.

@aik099 I think we should review only critical parts, all imrovements in docs, tests and code style should be accepted.

loren-osborn:
Closed by mistake. :-(

aik099:
> Gyus! ) Stop blaming PR and let's move to the merge part, it's impossible to do everything just in one single PR.

Agreed. I've already commented on each change and @loren-osborn is making necessary changes. As for replacing `parent(` and `static(` parts in `ReflectionClassLikeTrait` then this could be done in separate PR, because this PR already have lots of changes.

lisachenko:
So, @loren-osborn what could we do with this PR?

I really appreciate your contributions and want to accept your changes, so maybe we can split this single PR into several acceptable PRs? And just merge them one-by-one with respect to the @aik099 comments?

loren-osborn:
@lisachenko, like I said... let me revisit this late this afternoon. I don't see anything here really needing to be split up. (I already got a response from scrutinizer that they've opened tickets on the two false positives I reported.) Thanks again. Have a great evening.

src/ReflectionFile.php
```diff
@@ -50,6 +50,9 @@ class ReflectionFile
      */
     public function __construct($fileName, $topLevelNodes = null)
     {
+        if (!is_string($fileName)) {
```
aik099: It seems, that you've started to change more code. I'd recommend doing this in another PRs.

src/ReflectionFileNamespace.php
```diff
@@ -83,6 +83,9 @@ class ReflectionFileNamespace
      */
     public function __construct($fileName, $namespaceName, Namespace_ $namespaceNode = null)
     {
+        if (!is_string($fileName)) {
```
aik099: It seems, that you've started to change more code. I'd recommend doing this in another PRs.

src/ReflectionFile.php
loren-osborn: This was a change from my original commit from July 3rd before I ever issued this PR. I'm slightly confused.

src/ReflectionFileNamespace.php
loren-osborn: This was a change from my original commit from July 3rd before I ever issued this PR. I'm slightly confused.

src/bootstrap.php
```diff
@@ -12,10 +12,9 @@
...
+ * This file is used for automatic configuration of
+ * Go\ParserReflection\ReflectionEngine class
  */
 ReflectionEngine::init(new ComposerLocator());
-// Polifyll for PHP<7.0
-if (!class_exists(ReflectionType::class, false)) {
-    class ReflectionType {}
-}
+
+require(__DIR__ . '/polyfill.php');
```
loren-osborn: Hmm... I think this is a case of minor improvement in PSR-1 compliance for 1 extra small file load once, during execution. I think it's a worthwhile cost... and I think having Polyfills in a file called polyfill.php helps people know where to find them.

src/Traits/ReflectionClassLikeTrait.php
loren-osborn: marking with :+1: to remind me to revert.

src/ValueResolver/NodeExpressionResolver.php
loren-osborn: marking with :+1: to remind me to update.

src/ValueResolver/NodeExpressionResolver.php
loren-osborn: I agree we need to test for this case, but it isn't demonstrated in the example above. It would be something like this, which I'm doubting is valid PHP:

```php
const LEFT = 'Foo';
const RIGHT = 'Bar';

class FooBar {
    const FILENAME = 'baz.php';
}

require((LEFT . RIGHT)::FILENAME);
```

Like I said... I'm unsure if this is valid PHP, but I think, given that PhpParser accepts an expression, and we need something that can evaluate at compile time, something like this would be a valid test, even if not legal PHP currently.

Opinions?

loren-osborn:
> >  **An aside**: I mentioned several improvements I've made to this project (and Codeception/AspectMock) for a former employer that I wanted to share. I am no longer with this company, and they have declined any interest in contributing these in the foreseeable future, despite my urging. They did mention they **may** be willing to release these for a fee. No amount was discussed, but I suggested a hypothetical future employer of mine might want to pay "a few grand" (that's a few $1000 USD to non-native speakers) for such changes, that could save them several months of my salary to reproduce such features. I have no idea how receptive to this idea this company still is, but if any of you want to raise some funds for this, I can help you get in touch with those in a position to discuss releasing those changes.
>
> @loren-osborn , can you list the changes included in there, so that we'll know what's coming?

Hmm... The overarching changes are actually in Codeception/AspectMock:

- Adding the ability to Codeception/AspectMock to mock calls from code written in the default namespace
- Proper handling of functions and methods returning references, or taking references as arguments.
- Intelligent handling of error and exception handling, and custom autoloaders.
- Several other features that escape me at the moment.

This was all in support of running Drupal7 within AspectMock. Several changes were made to GoAOP!/ParserReflection to support this including:

- Recognizing and using inline parent classes/interfaces/traits without any special handling from the Locator. This was accomplished by bucket-sorting all functions, classes and constants when each file is parsed. (#73 and #64)
- Fixed some behavior inconsistent with native Reflections: (#72)
- Added passthrough to native reflection classes, so ParserReflections can be used for builtin classes. (#17)
- Added factory to support seamless reuse for derived reflection classes. (#2)
- Added support for multiple concurrent locators, so each locator fails over to the next.
- Added support for locators finding **functions** (and to some extent, constants) in addition to classes. Obviously this is not helpful for working with an autoloader (such as composer's), but can be appropriate for a custom locator.
- Added a "brute force" locator as a last-resort measure (to be used in situations where speed isn't a major concern)
- I'm sure there's more, but these are the enhancements that come to mind.

> Getting higher score on Scrutinizer CI won't be something worth reaching 😉 It's known for some weird rating calculating logic where some good classes are getting small ratings and attempt to fix rating results in making classes worse.
>
> In my projects I'm reviewing each of the errors and decide if that is even worth fixing or I can simply ignore that. Not sure if ignoring errors would improve my rating though. 😄

My goal wasn't to improve the Scrutinizer "score" per se. My goal was to make sure that all the issues were addressed. (Either but fixing the code, or determining that the issue didn't apply in our situation, and if appropriate, notify Scrutinizer of the false-positive... I have reported 2 from this PR.)

> @loren-osborn , review completed with outcome: don't trust Scrutinizer CI proposed fixes blindly 😉

@aik099,  I didn't even notice Scrutinizer had suggested patches until after I made most of the changes. Thank you for your very thorough review. Beginning to prepare a commit resolving most of these now.

src/ReflectionFileNamespace.php
aik099: Interesting. I've reviewed only changes that Scrutinizer CI reported as fixes to existing issues. Probably this change fixed nothing of Scrutinizer CI reported bugs.

In either case I think it's too much to check filename to be a string. Likely nobody would be crazy enough to pass non-string in there.

src/ReflectionFile.php
aik099: Interesting. I've reviewed only changes that Scrutinizer CI reported as fixes to existing issues. Probably this change fixed nothing of Scrutinizer CI reported bugs.

In either case I think it's too much to check filename to be a string. Likely nobody would be crazy enough to pass non-string in there.

src/ValueResolver/NodeExpressionResolver.php
aik099: Anything that would force PHP-Parser to create `Expr` class would do. Code won't be ever evaluated by PHP itself and only needs to be syntactically valid.

src/ReflectionFile.php
loren-osborn: The only reason I made this addition was that it was flagged by Scrutinizer:

> Documentation Bug introduced 30 days ago by @lisachenko
> It seems like `$fileName` can also be of type `array` or `Boolean`. However, the property `$fileName` is declared as type `string`. Maybe add an additional type check?

I think this based on type variations in the calling functions, despite this function only ever being called with string. I expect the other change that might fix this is isolating the parameter into a distinct, string-only variable that is used when calling this method. I think this type check is the easier solution.

src/ReflectionFileNamespace.php
loren-osborn: The only reason I made this addition was that it was flagged by Scrutinizer:

> Documentation Bug introduced 10 months ago by @lisachenko
> It seems like `$fileName` can also be of type `array` or `Boolean`. However, the property `$fileName` is declared as type `string`. Maybe add an additional type check?

(Same comment as above)

src/ReflectionFile.php
aik099: Then better to:

1. revert this change
2. in another PR address cases, when method can be called with non-string argument

src/ReflectionFileNamespace.php
aik099: Then better to:

1. revert this change
2. in another PR address cases, when method can be called with non-string argument

src/ReflectionFile.php
loren-osborn: The purpose of this PR was to address the Scrutinizer issues. This does that **and** it was in the PR from the beginning. 

I'm confused why this is suddenly an issue?

Let me investigate the calling code to see if that is more easily addressed than this. I'll get back to you.

aik099: 1. huge PRs are pain to review (this PR is becoming larger and larger)
2. the solution you've chosen for Scrutinizer CI reported problem isn't correct one

The Scrutinizer CI told that multiple types could come in and you've restricted to type check instead of actually investigating places from where multiple types could come in and deciding on fixing these places instead. This could be a lots of changes and that's why I've suggested another PR.

P.S.
You don't need to put ALL fixing in single PR. Better to send smaller manageable PRs. Right now this PR has large discussion that already is hard to track what's going on.

loren-osborn: Ok... I searched the code, and (outside of `tests/`) nothing within ParserReflection instantiates `Go\ParserReflection\ReflectionFile`, and ever use of it is annotated as `string`: The constructor is annotated `@param string`, the class property is annotated `@var string`, and the accessor is annotated `@return string`. I can only assume that this a plea for enforcing type safety rather than warning that badly typed data is being redirected here. I happen to be a fan of enforcing types where appropriate, so I see no harm in leaving this change, and the matching one for `ReflectionFileNamespace` intact.

👎 Keeping change in PR pending @lisachenko's approval.

src/ReflectionFileNamespace.php
loren-osborn: While there are 3 places in the code where `Go\ParserReflection\ReflectionFileNamespace` is instantiated, and two of them are called with filename coming from `Object(Go\ParserReflection\ValueResolver\NodeExpressionResolver)->context->getFileName()` where `context` isn't a strongly typed property, I believe the analysis from `Go\ParserReflection\ReflectionFile`'s constructor still holds: It's a good idea to do type checking where appropriate, and I think this can stay, unless @lisachenko wants me to yank it.

src/ValueResolver/NodeExpressionResolver.php
loren-osborn: Just for reference: [Variation of above sample code on 3v4l.org](https://3v4l.org/5i0aR)

src/ReflectionFile.php
aik099: Enforcing something, that will never happen would only slow down whole code checking process. For that reason I'd rather not check unless somebody stumbles upon this and reports it.

src/ReflectionMethod.php
loren-osborn: Ok... I confirmed that both cases are tested... (I did not flag the specific tests, but they're easy to find.) When `$returnType` is `null` the empty string value isn't used, but it prevents the `convertToDisplayType()` static method call that would be illegal with a `null` parameter. I am replacing the empty string with a string explaining its purpose.

loren-osborn:
@aik099, @lisachenko, I apologize dragging out this PR. I've had a rather busy and challenging past few days. I'll have this ready for a final review very soon.

loren-osborn:
@aik099, @lisachenko, I hope this is closer to what you were hoping for:

I left the filename string argument type check in place because scrutinizer had been complaining about it, but added tests for these cases. If you both agree I should back these out, I will.

I also added 2 more test files, as the cases I was trying to cover didn't seem to fit any of the existing files.

I believe the PR is ready for a second review. @lisachenko, can you weigh in on the filename type checking?

Thanks to both of you.

tests/ReflectionTypeTest.php
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Testing convertToDisplayType() with native \ReflectionType
+     *
+     * We're already testing it with Go\ParserReflection\ReflectionType
+     * elsewhere.
+     */
+    public function testTypeConvertToDisplayTypeWithNullableNativeType()
+    {
+        if (PHP_VERSION_ID < 70100) {
+            $this->markTestIncomplete('PHP >= 7.1 implements Nullable types');
```
aik099: If PHP 7.1+ implements nullable types, then test must be marked as skipped on other PHP versions, not an incomplete one.

tests/ReflectionTypeTest.php
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Testing convertToDisplayType() with native \ReflectionType
+     *
+     * We're already testing it with Go\ParserReflection\ReflectionType
+     * elsewhere.
+     */
+    public function testTypeConvertToDisplayTypeWithNativeType()
+    {
+        if (PHP_VERSION_ID < 70000) {
+            $this->markTestIncomplete('PHP >= 7.0 implements ReflectionType');
```
aik099: If PHP 7.0+ implements `ReflectionType`, then test must be marked as skipped on other PHP versions, not an incomplete one.

src/ReflectionMethod.php
```diff
@@ -80,7 +80,8 @@ public function ___debugInfo()
      */
     public function __toString()
     {
-        $hasReturnType    = $this->hasReturnType();
+        $returnType       = $this->getReturnType();
+        $hasReturnType    = ($returnType !== null); // Internally same as: $this->hasReturnType();
```
loren-osborn: It came up in our discussion, and it wasn't obvious to me, so I added the comment.

src/ReflectionMethod.php
```diff
@@ -107,14 +108,22 @@ public function __toString()
             $this->isFinal() ? ' final' : '',
             $this->isStatic() ? ' static' : '',
             $this->isAbstract() ? ' abstract' : '',
-            join(' ', \Reflection::getModifierNames($this->getModifiers() & 1792)),
+            join(
+                ' ',
+                \Reflection::getModifierNames(
+                    $this->getModifiers() &
+                        (self::IS_PUBLIC | self::IS_PROTECTED | self::IS_PRIVATE)
```
loren-osborn: Yeh... probably less relevant to this PR, but I generally hate non-expressive "magic numbers", and I can across this one while working on this PR. It probably doesn't belong, but I felt like it made the code clearer.

src/ReflectionMethod.php
```diff
@@ -107,14 +108,22 @@ public function __toString()
...
             $this->getName(),
             $this->getFileName(),
             $this->getStartLine(),
             $this->getEndLine(),
             count($methodParameters),
             $paramString,
-            $hasReturnType ? ReflectionType::convertToDisplayType($this->getReturnType()) : ''
+            (   $returnType ?
+                ReflectionType::convertToDisplayType($returnType) :
+                'UNUSED: Prevents convertToDisplayType() being called.')
```
loren-osborn: Ok, I guess my "not quite a comment" explanatory string was unclear. When `$returnType` is non-null, this is **definitely** used. When `$returnType` **is** `null`, however, the  `'%s'` substring that includes this in `sprintf()`'s return value, isn't included in the format string.  If you can think of a clearer way to say this, I welcome suggestions.

tests/ReflectionTypeTest.php
loren-osborn: Noted. I'll fix this. :+1:

tests/ReflectionTypeTest.php
loren-osborn: Agreed. My mistake. :+1:

loren-osborn: PHP 7.1+ implements **explicit** nullable types. I was considering including a 7.0+ test for implicit nullable types. It doesn't test a distinct code path, at least not in the code I was operating on here. Do you think I should add this test?

(By "implicit nullable types", I'm referring to parameters with class name type hints that default to `null`. I believe PHP 7.0+ is supposed to report those as `isNullable()`. I will double check this.)

src/ReflectionMethod.php
aik099: If `UNUSED: Prevents convertToDisplayType() being called.` text isn't actually displayed anywhere, then it doesn't make sense to write it.

If it's displayed, but same text isn't displayed in built-in reflection api, then this won't be compatible.

loren-osborn: It is indeed not displayed. I can easily change it back to the empty string. I'm trying to decide if I need a comment there to replace it.

I admit that from a clarity standpoint, I don't like the way this function reads, but I think anything but an empty string probably just muddies the water. @aik099, what do you think?

loren-osborn:
Just so you're aware, my laptop (which is my primary computer) is on the fritz so this may delay additional commits right now.

loren-osborn:
I changed the explicit compatibility checks to PHPUnit `@requires` annotations. I believe this is the last of the open issues.

@aik099, @lisachenko, Any other changes you want me to make before merging this in?

src/ReflectionFileNamespace.php
aik099: What is the 3rd place, can you post a link to that code line? If total call count in that 3rd place is much smaller than sum of other 2 places, then from performance point it would be better to type cast before calling this method instead of checking in constructor.

Same is valid for other similar place.

src/ReflectionMethod.php
aik099: I thought that you will change 

```php
$hasReturnType    = ($returnType !== null); // Internally same as: $this->hasReturnType();
```

into

```php
$hasReturnType    = $returnType !== null;
```

but you haven't.

src/ReflectionMethod.php
aik099: OK, let's keep it then.

src/ReflectionMethod.php
```diff
@@ -107,14 +108,20 @@ public function __toString()
             $this->isFinal() ? ' final' : '',
             $this->isStatic() ? ' static' : '',
             $this->isAbstract() ? ' abstract' : '',
-            join(' ', \Reflection::getModifierNames($this->getModifiers() & 1792)),
+            join(
+                ' ',
+                \Reflection::getModifierNames(
+                    $this->getModifiers() &
+                        (self::IS_PUBLIC | self::IS_PROTECTED | self::IS_PRIVATE)
```
aik099: Any way so solve this CS issue?

src/ReflectionMethod.php
```diff
@@ -107,14 +108,20 @@ public function __toString()
...
+                )
+            ),
             $this->getName(),
             $this->getFileName(),
             $this->getStartLine(),
             $this->getEndLine(),
             count($methodParameters),
             $paramString,
-            $hasReturnType ? ReflectionType::convertToDisplayType($this->getReturnType()) : ''
+            ($returnType ? ReflectionType::convertToDisplayType($returnType) : '')
```
aik099: Parenthesis around construct aren't required.

src/ReflectionType.php
```diff
@@ -85,7 +85,7 @@ public static function convertToDisplayType(\ReflectionType $type)
             'int'  => 'integer',
             'bool' => 'boolean'
         ];
-        $displayType = $type->type;
+        $displayType = (string) $type;
```
aik099: Are you sure, that according to coding standard there should be space between type name and variable?

src/ReflectionMethod.php
loren-osborn: That's fine... you want me to ditch the parenthesis? Can do...

src/ReflectionType.php
loren-osborn: I don't like the way it looks either... it's what Scrutinizer was suggesting. I'm happy to loose the ugly space. It's your call.

src/ReflectionMethod.php
loren-osborn: Adding the parentheses was inadvertent and I'll remove them, but I **do** think the comment is helpful. It's an explanation why we aren't calling `hasReturnType()`. @lisachenko seems to be deferring to you on this PR, so it's your call, but I think the comment has value.

if you disagree, I'll remove it; or if you prefer, I can put the comment on it's own line.

src/ReflectionMethod.php
loren-osborn: If I read the issue correctly, @Nitpick-CI wants `$this->getModifiers()` intended the same as `(self::IS_PUBLIC` which I was concerned may be easily mistaken as a second argument to `getModifierNames()`. Despite not liking overly long lines, I think getting rid of the line break is the best option.

As above, unambiguous options in my personal order of preference:

1. Remove the newline:
```php
            join(
                ' ',
                \Reflection::getModifierNames(
                    $this->getModifiers() & (self::IS_PUBLIC | self::IS_PROTECTED | self::IS_PRIVATE)
                )
            ),
```
2. Square up the two lines:
```php
            join(
                ' ',
                \Reflection::getModifierNames(
                    $this->getModifiers() &
                    (self::IS_PUBLIC | self::IS_PROTECTED | self::IS_PRIVATE)
                )
            ),
```
2. Leave them as-is:
```php
            join(
                ' ',
                \Reflection::getModifierNames(
                    $this->getModifiers() &
                        (self::IS_PUBLIC | self::IS_PROTECTED | self::IS_PRIVATE)
                )
            ),
```

src/ReflectionFileNamespace.php
loren-osborn: Ok... I'm happy to show you the three places that call this constructor, but the fact that you're asking indicates a failure to communicate on my part. I apologize for being unclear.

Here are the 3 places that call this constructor:


[.../parser-reflection/src/ReflectionFile.php](https://github.com/loren-osborn/parser-reflection/blob/295b2da02320cf978694172cd4ddc9a993cedfd8/src/ReflectionFile.php#L155):
```php
  153                  $namespaceName = $topLevelNode->name ? $topLevelNode->name->toString() : '';
  154  
  155:                 $namespaces[$namespaceName] = new ReflectionFileNamespace(
  156                      $this->fileName,
  157                      $namespaceName,
```

[.../parser-reflection/src/ValueResolver/NodeExpressionResolver.php](https://github.com/loren-osborn/parser-reflection/blob/295b2da02320cf978694172cd4ddc9a993cedfd8/src/ValueResolver/NodeExpressionResolver.php#L233):
```php
  231                  $fileName      = $this->context->getFileName();
  232                  $namespaceName = $this->resolveScalarMagicConstNamespace();
  233:                 $fileNamespace = new ReflectionFileNamespace($fileName, $namespaceName);
  234                  if ($fileNamespace->hasConstant($constantName)) {
  235                      $constantValue = $fileNamespace->getConstant($constantName);
```

[...](https://github.com/loren-osborn/parser-reflection/blob/295b2da02320cf978694172cd4ddc9a993cedfd8/src/ValueResolver/NodeExpressionResolver.php#L502)
```php
  500              $namespaceName = $this->resolveScalarMagicConstNamespace();
  501  
  502:             $fileNamespace = new ReflectionFileNamespace($fileName, $namespaceName);
  503              return $fileNamespace->getClass($className);
  504          }
```

That said, I think this largely misses the point. Let me see if I can clear this up a little bit:

- **First, an apology:** When you first brought these two changes up as issues, I think I took it a bit personally, and doubled down by adding unit tests for my type checking code. I was hoping that after trying to explain myself, why I made these changes would be clear, and I tried to defer to @lisachenko as an arbitrator. This basically blew up in my face, and I failed to address your concerns about these changes directly. I'm sorry. _Mea culpa._

- **Relevance:** The only reason I included these checks in this PR was because Scrutinizer flagged them as issues. While I try not to take anything that Scrutinizer says as gospel, in this case it has a point: These are public API entry points, and can be called with invalid arguments. Since we can't add PHP 7 type hints here without breaking PHP 5 compatibility, I felt it was appropriate to add these type checks to fix Scrutinizer's issue.

- **Performance:** You have repeatedly brought up the concern of this check impacting performance, but in these two instances it just isn't applicable. Of course I agree that any operation takes some amount overhead. The problem is that these are checks that only occur on the construction of infrequently created objects that **often** result in loading and parsing a file. Even if the parsed file cache is 80% effective, and we're pulling data from a solid state drive, or even the OS cache, checking the data type of a single function parameter has to be **at least** 3 to 5 orders of magnitude smaller reading and parsing a file 20% of the time. I agree that it contributes to constructor complexity which impacts maintainability, and both of those need to be kept in check, but this has **no appreciable** impact on performance.

- **Addressing this from where the method is called:** I think this is where our biggest misunderstanding lies.
    + My goal is to eliminate the issues being reported by Scrutinizer.
    + Scrutinizer is reporting two almost identical issues on two constructors with identically defined `$fileName` parameters.
    + One of those constructors is never called by this project's code, while the other is called three times.
    + **If** ensuring that all calling code was properly type hinted, then we should expect that the constructor that is never called (so all zero of its callers are properly type-hinted) would not trigger any Scrutinizer issue.
    + Since the uncalled constructor **does** trigger a Scrutinizer issue, I think it's safe to assume that ensuring the callers are properly type hinted (which I believe they already are) will have any impact on Scrutinizer reporting this issue.
    + **Also,** as this is a public API, I see no harm in a bit of sanity checking of parameters.

- **Summary:** The decision is ultimately yours, so if you ask me to get rid of these two checks, and the unit tests to support them, I will, but I think that asking that really misses the point of this PR.

**So, your choices are:**

1. I can leave these two `if (is_string())` checks, along with the corresponding unit tests, in place.
2. Or I can remove them.

loren-osborn:
@aik099, Thanks for the review. I'm ready to accept whatever changes you think are best. Just give me your final decisions on the following few alternatives: (I'm including my understanding of your current perference in parenthases.)

- `ReflectionFile` and `ReflectionFileNamespace` constructors: (Option 2: Remove the `is_string()` checks.)
- `ReflectionMethod` `$hasReturnType` comment: (Option 3: Strip the comment.)
- `ReflectionMethod` visibility bit mask: (Option 1: Remove the linebreak.)
- `ReflectionType` type cast whitespace: (Remove whitespace.)

While I hope I've been able to change your mind about the `is_string()` checks, I'll take your decisions as final.

Thanks again for all the time and attention to detail.

src/ReflectionType.php
aik099: Please locate other type cast operators in the code (not added by you) and see which style is used most. Then use most used style.

src/ReflectionMethod.php
aik099: Yes, please.

aik099:
> `ReflectionFile` and `ReflectionFileNamespace` constructors: (Option 2: Remove the is_string() checks.)

Nope. Please see https://github.com/goaop/parser-reflection/pull/76#discussion_r131549588 for actual idea.

> `ReflectionMethod` `$hasReturnType` comment: (Option 3: Strip the comment.)

Actually keeping. After pondering on this for a while I think comment here will help others to understand code better.

> ReflectionMethod visibility bit mask: (Option 1: Remove the linebreak.)

Whatever solves CS issue. I like the change in general.

> ReflectionType type cast whitespace: (Remove whitespace.)

Please locate other type cast operators in the code (not added by you) and see which style is used most. Then use most used style.

I hope my responses would allow you to make final changes to the PR.

loren-osborn:
@aik099, I'm rather confused. I'm happy to add explicit type casts to the three places `ReflectionFileNamespace` is instantiated if you like, but I'm fairly confident this won't address the issue. Please see my rather involved response to your comment.

Thanks

src/ReflectionFileNamespace.php
aik099: 1 - keep.

If we try to imagine real life examples, then `ReflectionFileNamespace` class would only be created once per namespace. And I haven't seen any codebases, where unique namespace count is fare superior (or even equal) to class/interface/trait count in these namespaces.

src/ReflectionFile.php
aik099: Let's keep it then. I hope this won't be a performance issue.

src/ReflectionFileNamespace.php
loren-osborn: Thank you for your quick response. I hope to have the final commit pushed this afternoon.

loren-osborn:
@aik099, would  you believe the whitespace after cast operators were split 50-50? There were only 4 casts total, so I made them all consistent by removing the whitespace from 2 existing casts. Everything else is what we discussed.

Thanks again.

aik099:
Merging, thanks @loren-osborn .

loren-osborn:
@aik099, Thanks again for all your hard work on this PR.

> > >  **An aside**: I mentioned several improvements I've made to this project (and Codeception/AspectMock) for a former employer that I wanted to share. I am no longer with this company, and they have declined any interest in contributing these in the foreseeable future, despite my urging. They did mention they **may** be willing to release these for a fee. No amount was discussed, but I suggested a hypothetical future employer of mine might want to pay "a few grand" (that's a few $1000 USD to non-native speakers) for such changes, that could save them several months of my salary to reproduce such features. I have no idea how receptive to this idea this company still is, but if any of you want to raise some funds for this, I can help you get in touch with those in a position to discuss releasing those changes.
> >
> > @loren-osborn , can you list the changes included in there, so that we'll know what's coming?
> 
> Hmm... The overarching changes are actually in Codeception/AspectMock:
> 
> - Adding the ability to Codeception/AspectMock to mock calls from code written in the default namespace
> - Proper handling of functions and methods returning references, or taking references as arguments.
> - Intelligent handling of error and exception handling, and custom autoloaders.
> - Several other features that escape me at the moment.
> 
> This was all in support of running Drupal7 within AspectMock. Several changes were made to GoAOP!/ParserReflection to support this including:
> 
> - Recognizing and using inline parent classes/interfaces/traits without any special handling from the Locator. This was accomplished by bucket-sorting all functions, classes and constants when each file is parsed. (#73 and #64)
> - Fixed some behavior inconsistent with native Reflections: (#72)
> - Added passthrough to native reflection classes, so ParserReflections can be used for builtin classes. (#17)
> - Added factory to support seamless reuse for derived reflection classes. (#2)
> - Added support for multiple concurrent locators, so each locator fails over to the next.
> - Added support for locators finding **functions** (and to some extent, constants) in addition to classes. Obviously this is not helpful for working with an autoloader (such as composer's), but can be appropriate for a custom locator.
> - Added a "brute force" locator as a last-resort measure (to be used in situations where speed isn't a major concern)
> - I'm sure there's more, but these are the enhancements that come to mind.
> 
> **Updated:** Just to clarify, none of this is "coming" unless someone can persuade my former employer (probably with a non-trivial monetary payment) to share these changes. 

I never did see any reaction from you to this list of changes. I realize that none of us is likely to be willing (or in my case able) to pay for these by themselves, but:

- Would a change list like this be welcome? (I realize getting a big heap of code cleaned up for integration into a project like this is a lot of work. I happen to know that @lisachenko wanted the cache of parsed files to remain shared as a class static, for instance, that I would need to change back.)
- I imagine if some of these are really wanted by users (I've seen repeated calls for global namespace support in AspectMock, for example) and a few dozen users want to pool their resources, it might be within reasonable reach? ($100-200/user split 20-35 ways? Perhaps I overestimate the interested user base?)
- If I can reproduce any of these features (gradually) here myself, would they be welcome?

aik099:
> Would a change list like this be welcome? (I realize getting a big heap of code cleaned up for integration into a project like this is a lot of work. I happen to know that @lisachenko wanted the cache of parsed files to remain shared as a class static, for instance, that I would need to change back.)

You can create each of mentioned improvements as an issue (lots of small manageable issues) and then anybody who is interested can express that interest.

I haven't personally encountered any of fixed issues or a need to described improvements.

> I imagine if some of these are really wanted by users (I've seen repeated calls for global namespace support in AspectMock, for example) and a few dozen users want to pool their resources, it might be within reasonable reach? ($100-200/user split 20-35 ways? Perhaps I overestimate the interested user base?)

If there is a platform for funding such type of work, then why not. I have no experience with cases like this (when former employer doesn't want to opensource changes for free).

> If I can reproduce any of these features (gradually) here myself, would they be welcome?

Time is not an issue here. If you report the issues/features as separate tasks, then you can eventually solve them.

loren-osborn:
@aik099, as was mentioned already, many of the improvements already have open issues for this project.

Oh, one other cool feature I didn't include in the above list (which is really more part of ApectMock, but I feel still relevant here) is ability to capture and intercept **some** custom autoloaders, overriding their `include`/`require` calls, and use them within ParserReflection's locators.

Issue #74, opened 2017-06-12
ReflectionClass::getParentClass with preloaded classes


MaximilianKresse:
I discovered a strange behaviour regardings classes which are already loaded / declared.

```php
<?php
// A.php
class A
{
    protected $test = 'Hello World';
}
```

```php
<?php
// B.php
class B extends A
{
}
```

```php
<?php
// index.php
require_once('vendor/autoload.php');

$locator = new \Go\ParserReflection\Locator\CallableLocator(function($className) {
    if (strpos($className, '\\') !== 0) {
        $className = '\\' . $className;
    }

    if ($className === '\A') {
        return __DIR__ . '/A.php';
    }
    
    if ($className === '\B') {
        return __DIR__ . '/B.php';
    }
    
    return false;
});
\Go\ParserReflection\ReflectionEngine::init($locator);

//require_once('A.php');

$parsedFile = new \Go\ParserReflection\ReflectionFile('B.php');
$fileNameSpaces = $parsedFile->getFileNamespaces();
foreach ($fileNameSpaces as $namespace) {
    $classes = $namespace->getClasses();
    foreach ($classes as $class) {
        foreach ($class->getProperties() as $property) {
            echo "Found class property: ", $class->getName(), '->', $property->getName(), ' default Value: ', $property->getValue(), PHP_EOL;
        }
    }
}
```

When you run this code the output is:
```
Found class property: B->test default Value: Hello World
```

But if I remove the comment before "require_once('A.php');" to load class A the output is:
```
Fatal error: Uncaught ReflectionException: Cannot access non-public member A::test
```

This problem only occurs when calling "getParentClass" (getProperties calls internally getParentClass). 
"getParentClass" returns a \ReflectionClass instead of a "\Go\ParserReflection\ReflectionClass". 

The problem seems to be here: https://github.com/goaop/parser-reflection/blob/master/src/Traits/ReflectionClassLikeTrait.php#L445

It could be fixed with:
```php
if (class_exists($extendsName, false)) {
    $parentClass = new parent($extendsName);
    if ($parentClass->isUserDefined()) {
        $parentClass = new static($extendsName);
    }
} else {
    $parentClass = new static($extendsName);
}
```

What do you think about this?

aik099:
> This problem only occurs when calling "getParentClass" (getProperties calls internally getParentClass).
"getParentClass" returns a \ReflectionClass instead of a "\Go\ParserReflection\ReflectionClass".

[...]

PR #77, opened 2017-08-18
Fixed getParentClass with preloaded Classes

Bob4ever:
See Issue #74

### DIFF
--- a/src/ReflectionClass.php
+++ b/src/ReflectionClass.php
@@ -137,6 +137,15 @@ class ReflectionClass extends InternalReflectionClass
      */
     protected function createReflectionForClass($className)
     {
-        return class_exists($className, false) ? new parent($className) : new static($className);
+        if (class_exists($className, false)) {
+            $parentClass = new parent($className);
+            if ($parentClass->isUserDefined()) {
+                $parentClass = new static($className);
+            }
+        } else {
+            $parentClass = new static($className);
+        }
+        
+        return $parentClass;
     }
 }


loren-osborn:
As I've mentioned before, I'm just a contributing user, so my opinions are just that: my opinions. While I created the`ReflectionClass::createReflectionForClass()` method, @lisachenko (and @aik099) will have their own ideas on how they want this project to grow. I added the`ReflectionClass::createReflectionForClass()` method as a temporary addition, and hoped to remove it in the next few PRs when the pass through @lisachenko proposed in #17 were implemented. I still think this is the preferred solution. Does anyone else care to weigh in?

aik099:
Would this change also fix #16 ?

Bob4ever:
@aik099 I think so. I didn't tested it but your issue seems to be the same.

aik099:
I think, that you should add tests to confirm, that correct class is used in all places, where returned object list contains mixed own/parent class properties/methods.

loren-osborn:
@aik099, would you want me to implement the pass through instead as an alternative? I think that is what @lisachenko said he preferred, and would a first step toward switching to a factory As discussed in #2.

aik099:
To keep PR focused just add tests for changes you've made and see if `getMethods` and `getProperties` methods would return proper (not built-in) reflection method/property instances for parent classes.

> would you want me to implement the pass through instead as an alternative?

Not sure that this PR is about that. If you'd like you can do that as well, but as separate PR. Honestly I'm not sure if factory idea from #2 is worth the effort spent on implementing it compared to how often people would use it.

loren-osborn:
I agree that it would be a separate PR, as would the factory, but a PR to merge **INSTEAD** of this one. The ParserReflection factory serves me very well when I used it with AspectMock previously.

aik099:
I prefer granular approach if time allows, where we make small (granular) changes and next changes can replace previous ones if needed.

Please do this and then in separate PR factory can be introduced:

> To keep PR focused just add tests for changes you've made and see if getMethods and getProperties methods would return proper (not built-in) reflection method/property instances for parent classes.

loren-osborn:
@aik099, after thinking more about this I've decided to break the factory feature changes into 3 PRs:

- One to add the pass through.
- Another to make ReflectionEngine's methods instance methods, and instantiate it.
- A third to implement the actual factory.

I'm still pondering a new project I'm contemplating adding as a dependency to the factory change to better deal with shared objects, but would appreciate you reviewing it when it's ready.

--------

With respect to **THIS** PR, my only concern is that it adds complexity to what was intended to be a temporary, throw-away method, mostly added to satisfy Scrutinizer.

aik099:
Do as you see fit then.

loren-osborn:
Just wanted to leave a progress report. This feature, with accompanying tests is almost complete. From my last commit:

> All tests pass again: added passthrough for Functions.
>
> Need to:
>   * Add ReflectionExtension code to emit proper classes.
>   * Refactor

loren-osborn:
I'm sure I seem like a royal flake by now. My 6 year old MacBook Pro died, and as I currently have no income, I'm hobbling by on my 13 year old HP that won't run anything newer than Windows XP. (I may switch it over to Linux soon, but right now I want to keep it **usable**. 

That said, I finally have a fully working implementation of the `ParserReflection` pas through that @lisachenko in #17.  I have a whole lot of refactoring I'd like to do, but with my current situation I'm working at a bit of a snails pace right now. 

While there is a good deal of refactoring I want to do to the main codebase once I've confirmed code coverage, but that should definitely go in a separate PR. The refactoring I want to do still in my pass through branch includes moving several of my utility test methods out of the test suite classes so they can be adequately tested. I haven't looked at much of this in most of a month, so I'm a little rusty, but I can't think of any other glaring issues.

I'm certainly open to any suggestions, but a new PR would be the right place for that. Anyone have a opinion if I should open a PR now, even if it's a bit unpolished, or wait until I think it will stand up better to scrutiny?

aik099:
You can open a new PR with `[WIP]` prefix in title to indicate that you want this to be reviewed, but it's not final version that can be merged.

lisachenko:
I'm going to close this in favor of #82

PR #78, opened 2017-08-27
Configure lexer to capture token positions for new parser

lisachenko:
Token position capturing will be used Go! AOP for AST-transformation and can simplify processing


lisachenko:
Linked to https://github.com/goaop/framework/pull/362

PR #81, opened 2017-09-24
[WIP] Native reflection pass-through

loren-osborn:
This patch allows ParsedReflections to fall back to native implementation when dealing with built-in functions that have no parsable source code. This was proposed by @lisachenko in #17.

I think this implementation is complete, but it's still a bit rough. I especially want to move the utility functions I use in my tests into their own class(es) and give them their own tests. `Go\ParserReflection\TestCaseBase::getStringificationOf()` is one such bit of nastiness that should have it's own tests.

Thanks for any feedback.

### DIFF
--- a/src/IReflection.php
+++ b/src/IReflection.php
@@ -9,7 +9,7 @@
  */
 namespace Go\ParserReflection;
 
-interface IReflector
+interface IReflection
 {
     public function wasIncluded();
 }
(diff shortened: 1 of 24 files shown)


loren-osborn:
With all the scrutinizer issues and changes, and my 2 dozen or so commits already, this is a very noisy PR. I will squash the commits and submit a new PR later today.

PR #82, opened 2017-09-24
[WIP] Native reflection pass-through

loren-osborn:
(This is a reissue of #81 to hide initial cleanup noise.)

This patch allows ParsedReflections to fall back to native implementation when dealing with built-in functions that have no parsable source code. This was proposed by @lisachenko in #17.

I think this implementation is complete, but it's still a bit rough. I especially want to move the utility functions I use in my tests into their own class(es) and give them their own tests. `Go\ParserReflection\TestCaseBase::getStringificationOf()` is one such bit of nastiness that should have it's own tests.

Thanks for any feedback.


loren-osborn:
@Bob4ever, if we go in this direction, this will probably obsolete #77.

loren-osborn:
My schedule is still pretty unpredictable for the time being, but @lisachenko, can you give me a sense of how close this is to what you had envisioned for an API passthrough? Do you see anything particular you might want me to change? I know this is a pretty big patch, and I certainly don't expect a code review on something still in progress. I'm simply wanting your overall impressions.

Thanks

lisachenko:
Wow! It's a huge job! Thanks!

From my first quick view, it's what I planned to do previously, need to look deeper into PR.

tests/TestCaseBase.php
```diff
@@ -0,0 +1,306 @@
...
+     *
+     * @return string
+     */
+    protected function getShortNameFromName($name)
+    {
+        $nameParts = explode('\\', $name);
+        return array_pop($nameParts);
+    }
+
+    protected function getStringificationOf($value, $maxLen = 128)
```
lisachenko: Wow! Why do we need such a complex function? I'd vote to drop it and replace with something more general, without dark magic inside...

src/IReflection.php
```diff
@@ -0,0 +1,15 @@
...
+ * Parser Reflection API
+ *
+ * @copyright Copyright 2015, Lisachenko Alexander <lisachenko.it@gmail.com>
+ *
+ * This source file is subject to the license that is bundled
+ * with this source code in the file LICENSE.
+ */
+namespace Go\ParserReflection;
+
+interface IReflection
```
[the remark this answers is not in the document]
loren-osborn: The purpose of this interface is intended to be more than it currently appears to be. Because we don't have multiple inheritance the idea was that each Reflection class would have an `IReflection...` interface, and that each of these would extend the parent of the original Reflection class. The purpose being that you could easily type-hint any parser-reflection class, with all compatible subtypes by using, for instance `IReflectionFunctionAbstract`, which would be implemented by all subtypes that are parser-reflection classes. I thought a full "Interface" prefix for this purpose would become too cumbersome. Also, native Reflections already have a `Reflector` interface, so, if I didn't want to go in this direction, I probably would have named this `Reflector` instead.  I also didn't want the compatibility headaches of renaming the interface later.

Interested to hear your feedback.

src/IReflection.php
```diff
@@ -0,0 +1,15 @@
...
+{
+    public function wasIncluded();
```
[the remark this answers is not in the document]
loren-osborn: The Interface I explained above. The method was something I felt was intrinsic to the feature that parser-reflection classes offer: The ability to Reflect a construct **before** it's included (any built-in construct is already "included"). I believe I do use this in part of my of my passthrough implementation, but I thought knowing if a construct was already defined was intrinsic to what a parser-reflection class is.

This is also why I think adding a factory mechanism, which this patch is the first step towards, (and you indicated that you are sitting on the fence about) will be such a powerful feature. It allows seemlessly layering these application specific derived classes into the existing ecosystem of reflection classes. 

I look forward to your thoughts.

src/ReflectionFunction.php
```diff
@@ -18,25 +18,46 @@
...
         // Remove the last one part with function name
         array_pop($namespaceParts);
         $this->namespaceName = join('\\', $namespaceParts);
-
         $this->functionLikeNode = $functionNode;
+        if (!$this->functionLikeNode) {
+            $isUserDefined = true;
+            if ($this->wasIncluded()) {
+                $nativeRef = new BaseReflectionFunction($functionName);
+                $isUserDefined = $nativeRef->isUserDefined();
```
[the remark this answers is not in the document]
loren-osborn: Wasn't sure if you had a preference there. I'm happy to fix. Not always sure when applicable, but I'll see what I can do.

src/ReflectionParameter.php
```diff
@@ -217,6 +222,7 @@ public function getClass()
 
     /**
      * {@inheritDoc}
+     * @return Go\ParserReflection\ReflectionClass|null The class of the method that declared the parameter, if any.
```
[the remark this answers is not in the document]
loren-osborn: I actually prefer aliased class names, but I was trying to follow style guidelines (or possibly PHPCodeSniffer) which insisted on FQDNs everywhere. This change was mostly for Scrutinizer's benefit, so if it understands class aliases in doc-blocks, I whole-heartedly agree.

If Scrutinizer can't grok them, should I still make this change?

src/ReflectionFunction.php
```diff
@@ -136,6 +161,17 @@ public function __toString()
      */
     protected function __initialize()
     {
-        parent::__construct($this->getName());
+        parent::__construct($this->functionName);
+    }
+
+    /**
+     * Has class been loaded by PHP.
```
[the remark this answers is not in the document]
loren-osborn: Can you give me an example of the proper format? What did I miss? -Thanks.

src/ReflectionProperty.php
```diff
@@ -60,13 +68,31 @@ public function __construct(
         Property $propertyType = null,
         PropertyProperty $propertyNode = null
     ) {
-        $this->className    = $className;
+        $this->className        = $className;
+        $this->propertyName     = $propertyName;
         if (!$propertyType || !$propertyNode) {
-            list ($propertyType, $propertyNode) = ReflectionEngine::parseClassProperty($className, $propertyName);
+            $oneNodeProvided = $propertyType || $propertyNode;
```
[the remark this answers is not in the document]
loren-osborn: No, in this case I was specifically referring to the quantity 1.  This is to catch an invalid case, as neither or both of these parameters should be non-null. This would probably be more readable if I changed it to an `xor` operator.

src/Traits/ReflectionClassLikeTrait.php
```diff
@@ -346,19 +393,31 @@ public function getMethod($name)
...
+                });
+                $methods = $directMethods + $parentMethods;
 
-            $this->methods = $methods;
+                $this->methods = $methods;
+            } else {
+                $this->initializeInternalReflection();
+                $methodRefs = parent::getMethods();
+                $this->methods = [];
+                foreach ($methodRefs as $eachMethod) {
```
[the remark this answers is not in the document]
loren-osborn: 👍 Sounds good. I wanted to avoid using `$methodRefs` with `$methodRef`, as that would be rather hard to distinguish.

src/Traits/ReflectionFunctionLikeTrait.php
```diff
@@ -339,6 +436,80 @@ public function isVariadic()
...
     }
+
+    private function getRefParam(BaseReflectionParameter $orig)
+    {
+        $nullableImplied = false;
+        $builder = new ParamNodeBuilder($orig->name);
+        if ($orig->isDefaultValueAvailable() || $orig->isOptional()) {
+            if ($orig->isDefaultValueAvailable()) {
+                if (method_exists($orig, 'isDefaultValueConstant') && $orig->isDefaultValueConstant()) {
+                    $constNameParts = explode('::', $orig->getDefaultValueConstantName(), 2);
```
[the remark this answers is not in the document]
loren-osborn: I'm fine with moving part of this to `NodeExpressionResolver`,  but it seems built to process nodes rather than strings. How would you recommend I delegate this to the parser? My gut reaction would be to give it something like:

```php
sprintf('<?php function foo($bar = %s) {}', $orig->getDefaultValueConstantName())
```

then pluck out the resulting AST nodes for the default value. Is there a cleaner way?

tests/AbstractClassTestCaseBase.php
```diff
@@ -82,6 +87,40 @@ public function getFilesToAnalyze()
...
+        $classes = [];
+        foreach ($builtInClasses as $className) {
+            $classes[$className] = ['class' => $className, 'fileName'  => null];
+        }
+        $files = $this->getFilesToAnalyze();
+        foreach ($files as $filenameArgList) {
+            $argKeys = array_keys($filenameArgList);
+            $fileName = $filenameArgList[$argKeys[0]];
+            $resolvedFileName = stream_resolve_include_path($fileName);
+            $fileNode = ReflectionEngine::parseFile($resolvedFileName);
```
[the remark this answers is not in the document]
loren-osborn: I'll look for a Sublime plugin that does that... what IDE(s)/editor(s) do you use?

tests/TestCaseBase.php
loren-osborn: This method is **most** of why this PR is still `[WIP]` right now... it is there exclusively for test failure readability, so I think it adds minimal risk, but needs its own unit tests, and should be moved into a class in `tests/Support`, or somewhere similar.

Opinions?

src/IReflection.php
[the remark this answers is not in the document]
loren-osborn: ...or perhaps you want me to rename this `Reflector`,  and add an empty derived interface (`IReflection` or `ReflectionInterface` or `Interfaces\Reflection` or whatever you think is best) when I add all its brethren? (Kicking the van down the road so to speak.)

tests/TestCaseBase.php
loren-osborn: Let me amend that.... I intend to use this function for future tests, not just test failure reporting, but it is **currently** only used for test error reporting.

src/IReflection.php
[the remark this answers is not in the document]
lisachenko: My POV is to name it as `ParserReflection\ReflectionInterface` and leave all stuff in this interface

src/IReflection.php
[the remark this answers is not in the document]
lisachenko: I like your explanation, so let's go with `wasIncluded` method in the interface.

src/ReflectionFunction.php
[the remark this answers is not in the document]
lisachenko: Maybe it's my personal taste but when I'm looking at aligned code it reads more natural and simpler.

src/ReflectionFunction.php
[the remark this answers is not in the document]
lisachenko: This method relates to the function, so it should be related to the function and not to the class :) So, naming should be "Has function/method been loaded by PHP"

src/ReflectionParameter.php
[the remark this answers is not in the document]
lisachenko: From what I know, Scrutinizer fully understands aliased class names, so we can use aliases in code base.

src/ReflectionProperty.php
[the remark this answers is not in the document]
lisachenko: Can't agree with you. Because when both parameter are present, then this value will be `true`. So it should be named `atLeastOneNode`. For exactly one node it should be calculated as `($propertyType && !$propertyNode) || (!$propertyType && $propertyNode)`

So, still need to fix naming or expression to match.

src/IReflection.php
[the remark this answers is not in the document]
loren-osborn: 👍 

Still tempted to add `Go\ParserReflection\Reflector` in the future for consistency, but I see no need for that in this PR.

src/IReflection.php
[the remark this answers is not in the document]
loren-osborn: 👍

src/ReflectionFunction.php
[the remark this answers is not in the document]
loren-osborn: I prefer it too, but we might want a rule of thumb to ad consistency. (i.e. Min number of consecutive assignments to align. How to mix with broken lines. Should they cascade into nested blocks. etc.)

For now I will only align 2 or more consecutive assignments with no intervening broken lines unless you have a more specific preference.

src/ReflectionFunction.php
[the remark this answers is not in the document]
loren-osborn: That looks like a typo/copy-paste error... I've been using the name "construct" to refer to functions/classes/constants/variables, but I'm open to a better name.

src/ReflectionParameter.php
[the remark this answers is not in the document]
loren-osborn: Either way is fine. I **DO** want to double check to see if there's a PSR recommendation on this. I don't mind defying it, but I prefer to do so **knowingly**.  (If you want me to defer to the PSR recommendation, if I find one, I can do that instead if you like.)

src/ReflectionProperty.php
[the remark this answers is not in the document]
loren-osborn: It's possible that my logic is broken, but we should only be in this block if one or both values are null (or falsey), so they can't both be non-null. (This may be confusing because this happens before the assignment which is in red above it.) I will change to `xor` for clarity.

lisachenko:
Hi, @loren-osborn! Sorry for the delay, was on vacation.

Could we proceed with this PR?

I also want to ask you and @aik099 if you want to become library developer with rights to push changes into master branch.

src/Traits/ReflectionFunctionLikeTrait.php
```diff
@@ -63,33 +70,59 @@ public function getClosureThis()
...
-        return null;
+        $extName = $this->getExtensionName();
+        if (!$extName) {
+            return null;
+        }
+        // The purpose of Go\ParserReflection\ReflectionExtension is
+        // to behave exactly like \ReflectionExtension, but return
+        // Go\ParserReflection\ReflectionFunction and
+        // Go\ParserReflection\ReflectionClass where apropriate.
+        return new ReflectionExtension($extName);
```
aik099: Is it tested?

To actually test it you need to know what extension are loaded into PHP and try to reflect class provided by them.

src/ReflectionExtension.php
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Returns list of reflection classes
+     *
+     * @return array|\ReflectionClass[]
+     */
+    public function getClasses()
+    {
+        if (!isset($this->classes)) {
+            $classRefs = parent::getClasses();
```
aik099: You can safely inline this variable, because it's used only once.

src/ReflectionExtension.php
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Returns list of reflection functions
+     *
+     * @return array|\ReflectionFunction[]
+     */
+    public function getFunctions()
+    {
+        if (!isset($this->functions)) {
+            $funcRefs = parent::getFunctions();
```
aik099: You can safely inline this variable, because it's used only once.

src/ReflectionExtension.php
```diff
@@ -0,0 +1,76 @@
...
+     * Returns list of reflection classes
+     *
+     * @return array|\ReflectionClass[]
+     */
+    public function getClasses()
+    {
+        if (!isset($this->classes)) {
+            $classRefs = parent::getClasses();
+            $this->classes = [];
+            foreach ($classRefs as $origKey => $eachClass) {
```
aik099: Please rename:

* `$origKey` into `$index`
* `$eachClass` into `$classReflection`

---

I think rename makes names easier to understand.

src/ReflectionExtension.php
```diff
@@ -0,0 +1,76 @@
...
+     * Returns list of reflection functions
+     *
+     * @return array|\ReflectionFunction[]
+     */
+    public function getFunctions()
+    {
+        if (!isset($this->functions)) {
+            $funcRefs = parent::getFunctions();
+            $this->functions = [];
+            foreach ($funcRefs as $origKey => $eachFunc) {
```
aik099: Please rename:

* `$origKey` into `$index`
* `$eachFunc ` into `$functionReflection`

---

I think rename makes names easier to understand.

src/ReflectionExtension.php
```diff
@@ -0,0 +1,76 @@
...
+     */
+    public function wasIncluded()
+    {
+        return true;
+    }
+
+    /**
+     * Returns list of reflection classes
+     *
+     * @return array|\ReflectionClass[]
```
aik099: 1. Please remove `array|`, because `[]` ending of `\ReflectionClass[]` already indicates to IDE that it's an array.
2. Shouldn't `\ReflectionClass[]` be `ReflectionClass[]` (namespaced version) of class instead?

src/ReflectionExtension.php
```diff
@@ -0,0 +1,76 @@
...
+            }
+        }
+
+        return $this->classes;
+    }
+
+    /**
+     * Returns list of reflection functions
+     *
+     * @return array|\ReflectionFunction[]
```
aik099: 1. Please remove `array|`, because `[]` ending of `\ReflectionFunction[]` already indicates to IDE that it's an array.
2. Shouldn't `\ReflectionFunction[]` be `ReflectionFunction[]` (namespaced version) of class instead?

src/ReflectionFunction.php
```diff
@@ -18,25 +18,46 @@
...
+    private $functionName;
+
+    /**
      * Initializes reflection instance for given AST-node
      *
      * @param string|\Closure $functionName The name of the function to reflect or a closure.
      * @param Function_|null  $functionNode Function node AST
      */
-    public function __construct($functionName, Function_ $functionNode)
+    public function __construct($functionName, Function_ $functionNode = null)
```
aik099: 1. When the `$functionNode` would be missing?
2. Why `$functionNode` wasn't missing before?

src/ReflectionFunction.php
```diff
@@ -18,25 +18,46 @@
...
         $this->namespaceName = join('\\', $namespaceParts);
-
         $this->functionLikeNode = $functionNode;
+        if (!$this->functionLikeNode) {
+            $isUserDefined = true;
+            if ($this->wasIncluded()) {
+                $nativeRef = new BaseReflectionFunction($functionName);
+                $isUserDefined = $nativeRef->isUserDefined();
+            }
+            if ($isUserDefined) {
```
aik099: How is `user defined` correlate with `!$this->functionLikeNode`?

src/ReflectionFunction.php
```diff
@@ -136,6 +161,17 @@ public function __toString()
...
+
+    /**
+     * Has class been loaded by PHP.
+     *
+     * @return bool
+     *     If class file was included.
+     */
+    public function wasIncluded()
+    {
+        return function_exists($this->functionName);
```
aik099: Theoretically this is possible, but do we handle case with functions auto-loaded (if that's even possible)?

src/ReflectionFunction.php
```diff
@@ -18,25 +18,46 @@
...
+class ReflectionFunction extends BaseReflectionFunction implements IReflection
 {
     use ReflectionFunctionLikeTrait, InternalPropertiesEmulationTrait;
 
     /**
+     * Name of the function
+     *
+     * @var string|\Closure
+     */
+    private $functionName;
```
aik099: Any benefits of remembering user-provided `$functionName` VS using `\ReflectionFunction::getName()`?

src/ReflectionMethod.php
```diff
@@ -38,6 +39,13 @@ class ReflectionMethod extends BaseReflectionMethod
     private $declaringClass;
 
     /**
+     * Name of method
+     *
+     * @var string
+     */
+    private $methodName;
```
aik099: Any benefits of remembering user-provided `$methodName` VS using `\ReflectionMethod::getName()`?

src/ReflectionMethod.php
```diff
@@ -224,6 +250,10 @@ public function invokeArgs($object, array $args)
      */
     public function isAbstract()
     {
+        if (!$this->functionLikeNode) {
```
aik099: When this is possible?

Same comment applies to other methods in the class.

src/ReflectionParameter.php
```diff
@@ -89,7 +89,7 @@ public function __construct(
         $this->parameterIndex    = $parameterIndex;
         $this->declaringFunction = $declaringFunction;
 
-        if ($this->isDefaultValueAvailable()) {
+        if ($this->isDefaultValueSet()) {
```
aik099: Why this is changed?

src/ReflectionParameter.php
```diff
@@ -165,6 +165,11 @@ public function __toString()
      */
     public function allowsNull()
     {
+        // Allow builtin types to override
+        if ($this->parameterNode->getAttribute('prohibit_null', false)) {
```
aik099: Does this belong to this PR or should a separate bugfix PR with connected unit tests?

src/ReflectionParameter.php
```diff
@@ -329,7 +336,9 @@ public function isCallable()
      */
     public function isDefaultValueAvailable()
     {
-        return isset($this->parameterNode->default);
+        return
+            isset($this->parameterNode->default) &&
```
aik099: Does this belong to this PR or should a separate bugfix PR with connected unit tests?

src/ReflectionParameter.php
```diff
@@ -345,7 +354,7 @@ public function isDefaultValueConstant()
      */
     public function isOptional()
     {
-        return $this->isVariadic() || ($this->isDefaultValueAvailable() && $this->haveSiblingsDefalutValues());
+        return $this->isVariadic() || ($this->isDefaultValueSet() && $this->haveSiblingsDefalutValues());
```
aik099: Does this belong to this PR or should a separate bugfix PR with connected unit tests?

src/ReflectionParameter.php
```diff
@@ -365,6 +374,23 @@ public function isVariadic()
     }
 
     /**
+     * Returns if default value set (or implied).
+     *
+     * Identical to isDefaultValueAvailable(), except it
+     * includes IMPLIED default values which:
+     *     + Only exist in builtin functions and methods.
+     *     + Only affect the optionality of a prameter, not
```
aik099: Typo here in `prameter` word.

src/ReflectionParameter.php
```diff
@@ -365,6 +374,23 @@ public function isVariadic()
...
+     *         if a SUPPLIED parameter can have the default
+     *         value.
+     *
+     * @return bool
+     */
+    protected function isDefaultValueSet()
```
aik099: 1. Any cases, when `isDefaultValueSet` would work better, than calling `isDefaultValueAvailable`?
2. Maybe instead we should improve `isDefaultValueAvailable`?

src/ReflectionProperty.php
```diff
@@ -98,7 +124,7 @@ public function getTypeNode()
     public function ___debugInfo()
     {
         return array(
-            'name'  => isset($this->propertyNode) ? $this->propertyNode->name : 'unknown',
+            'name'  => isset($this->propertyName) ? $this->propertyName : 'unknown',
```
aik099: The `isset` won't get you desired result, because `$this->propertyName` is now always set (thanks to your changes) from mandatory constructor argument.

src/Traits/ReflectionFunctionLikeTrait.php
```diff
@@ -53,7 +60,7 @@ public function getClosureScopeClass()
     {
         $this->initializeInternalReflection();
 
-        return forward_static_call('parent::getClosureScopeClass');
+        return parent::getClosureScopeClass();
```
aik099: Why change this?

src/Traits/ReflectionFunctionLikeTrait.php
```diff
@@ -63,33 +70,59 @@ public function getClosureThis()
     {
         $this->initializeInternalReflection();
 
-        return forward_static_call('parent::getClosureThis');
+        return parent::getClosureThis();
```
aik099: Why change this?

src/Traits/ReflectionFunctionLikeTrait.php
```diff
@@ -63,33 +70,59 @@ public function getClosureThis()
...
     }
 
     public function getDocComment()
     {
+        if (!$this->functionLikeNode) {
```
aik099: I see lots of similar checks for internal reflection initialization. Maybe we can create a proxy classes that would use `__call` method where optional initialization of internal reflection would be made.

That would greatly reduce copy/paste of such code, but I'm not sure if excessive usage of `__call` method would result in performance decrease.

src/Traits/ReflectionFunctionLikeTrait.php
```diff
@@ -181,18 +230,32 @@ public function getParameters()
...
+            if (is_object($returnType)) {
+                $returnType = $returnType->toString();
+            } elseif (is_string($returnType)) {
+                $isBuiltin = true;
+            } else {
+                return null;
+            }
         } else {
-            return null;
+            if (PHP_VERSION_ID < 70000) {
```
aik099: This check won't help you, because we can use PHP5 to parse PHP7 files and vice-versa.

loren-osborn:
Ok... just dashing off a quick note... I'd fallen off the radar a bit myself for a bit, and hobbling by with a Windows XP laptop at the moment, so somewhat underequipped to test any PHP 7 code at the the moment, except on my iMac at home. I do have a bunch more PRs I want to contribute, but had to focus on other areas of my life recently.

@lisachenko, I greatly appreciate the offer to be a library contributor, and am happy to help any way I can. Prudence dictates that I should still rely on your very helpful peer reviews, if that isn't inconvienient.

@aik099, thanks for the second round of review comments. I will respond as soon as I am able.

Give me a day or two to get back up to speed on this PR.

aik099:
@loren-osborn , sure, take your time. Whenever you're ready just add a comment so that I can know that push/commit round was made a ready to be reviewed.

src/ReflectionParameter.php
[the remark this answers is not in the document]
loren-osborn: I consulted PSR-5 and confirmed that you are correct. Fully qualified class names (FQCNs), namespace-relative class names and aliased class names are all valid to use in docblock for type specification. The relevant text is in [PSR-5, Appendix A. Section 3](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#valid-class-name). As you suggested I will switch these to aliased or namespace-relative class names.

loren-osborn:
@aik099, based on your review comments, I wanted to give some overall context for this PR that may be unclear:

The goal of this PR is to bring feature pairity to `Go\ParserReflection` classes with their native equivalents for functions/classes/methods/etc. that are **compiled into PHP** (and therefore have no parsable PHP source code.) Because these functions/classes/methods/etc. previously weren't supported by `Go\ParserReflection` classes, any interaction (such as inheritance) between built in and user defined functions/classes/methods/etc. needed to be special cased. As this PR makes this no longer the case, many of these special cases disappear.

Of course, we don't get this new flexibility for free, so we have lots of **new** special cases for most `Go\ParserReflection` classes, for cases when there is no PHP source code. In these cases, parsed node properties, like `$this->functionLikeNode` and `$this->classLikeNode`, are `null`.

As far as testing all of this new functionality, I did add tests for these cases, in general, but I need to go back and check each of the conditions you pointed out individually, which I fully intend to do. I will use a :memo: to remind myself of tests I need to go back and verify, as well as pointing out any tests that I am already aware are missing/insufficient.

These are simply overall responses to the issues I saw you raise. I'll go back now and respond to your individual points. These are the outstanding issues I'm already aware of, but I appreciate any feedback you have on these as well:

### Known Outstanding Issues:

* Already pointed out to @lisachenko:
     - [ ] Pull non-testing code out of test classes.
     - [ ] Put this test-support code under test.
     - [ ] Refactor this test-support code.
* Previously undiscussed issues:
     - [ ] Add ability to construct `Go\ParserReflection\ReflectionParameter` with `string` and `string|int` arguments respectively`$function` and `$parameter` like native `ReflectionParameter`.
* Issues for separate PR:
     - [ ] Major refactor, especially of test code, (hopefully) bringing 100% code coverage.

src/ReflectionClass.php
```diff
@@ -41,7 +41,17 @@ public function __construct($argument, ClassLike $classLikeNode = null)
...
 
-        $this->classLikeNode = $classLikeNode ?: ReflectionEngine::parseClass($fullClassName);
+        $this->classLikeNode = $classLikeNode;
+        if (!$this->classLikeNode) {
+            $isUserDefined = true;
+            if ($this->wasIncluded()) {
+                $nativeRef = new BaseReflectionClass($fullClassName);
+                $isUserDefined = $nativeRef->isUserDefined();
+            }
+            if ($isUserDefined) {
```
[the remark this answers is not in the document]
loren-osborn: Looks like this code is much less clear than I realized. I will wrap this logic into a new private method `isParsedNodeMissing()`. I believe the tests I added cover this new case, but I will verify this. :memo: To understand why this new logic is necessary, let me contrast the old and new behavior:

### The Old Behavior
* If the parsed `Node` wasn't supplied:
    * Have the `ReflectionEngine` find it using the locator
        * If the locator can't find the class source code, throw a (hopefully fatal) exception.

Now we need to support classes that may not have PHP source code. Keep in mind "user defined" simply means PHP source code exists. So I changed this to:

### The New Behavior
* If no parsed `Node` is supplied:
    * The class is considered user defined unless the class is already defined in PHP.
    * If the class is already defined in PHP:
        * The class is user defined if native `ReflectionClass` says it is.
    * If the class is user defined:
        * Ask `ReflectionEngine` to load the PHP using the locator as before.

This way we aren't searching for PHP source files for built in classes. Does this clarify the new logic at all?

src/ReflectionClass.php
```diff
@@ -62,11 +72,8 @@ public static function collectInterfacesFromClassNode(ClassLike $classLikeNode)
         if ($implementsList) {
             foreach ($implementsList as $implementNode) {
                 if ($implementNode instanceof FullyQualified) {
-                    $implementName  = $implementNode->toString();
-                    $interface      = interface_exists($implementName, false)
-                        ? new parent($implementName)
-                        : new static($implementName);
-                    $interfaces[$implementName] = $interface;
+                    $implementName              = $implementNode->toString();
+                    $interfaces[$implementName] = new static($implementName);
```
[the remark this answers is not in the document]
loren-osborn: 1. It used to be needed, because `Go\ParserReflection\ReflectionClass` didn't previously support built in classes (or interfaces in this case), but now it does.
1. This new situation **should** already be under test, but I will verify: :memo:

src/ReflectionClass.php
```diff
@@ -92,10 +99,7 @@ public static function collectTraitsFromClassNode(ClassLike $classLikeNode, arra
                     foreach ($classLevelNode->traits as $classTraitName) {
                         if ($classTraitName instanceof FullyQualified) {
                             $traitName          = $classTraitName->toString();
-                            $trait              = trait_exists($traitName, false)
-                                ? new parent($traitName)
-                                : new static($traitName);
-                            $traits[$traitName] = $trait;
+                            $traits[$traitName] = new static($traitName);
```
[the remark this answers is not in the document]
loren-osborn: (See last answer)

src/ReflectionExtension.php
loren-osborn: I will look into and play around with this. My recollection is that's the return type Scrutinizer extracted from the method from the parent class like the property initialized to an empty array of the type of it's members was specified. I don't recall if the issue came up here, or with the calling code. I will investigate further. :+1:

src/ReflectionExtension.php
loren-osborn: I'll try this. 👍

src/ReflectionExtension.php
loren-osborn: Again, I'll try to fix this without Scrutinizer throwing a tantrum. 👍

src/ReflectionFunction.php
loren-osborn: `parent::getName()` is not initialized in all cases. Less to special case.

src/ReflectionFunction.php
loren-osborn: `$functionNode` will be missing when `$functionName` names a built in function. (When there is no PHP source code to parse.)

In the future, I expect the `ReflectionEngine` to remember where it found named functions, but not in this PR.

src/ReflectionFunction.php
loren-osborn: For a fully constructed `ReflectionFunction` object,  `$this->isUserDefined()` should always return `!$this->functionLikeNode`.  I believe I am actually returning the equivalent of `!($this->functionLikeNode || !parent::isUserDefined())` just in case there some condition I didn't account for, but too my knowledge these should always be the same.

src/ReflectionFunction.php
loren-osborn: To my knowledge, PHP function autoloading has been proposed several times, but never included in any mainstream branch. If function auto loading were implemented, and `function_exists()` had the same semantics as `class_exists()` we would need to add a `false` as a second parameter, but that's simply conjecture at this point.

src/ReflectionMethod.php
loren-osborn: (As above:) `parent::getName()` is not initialized in all cases. Less to special case.

src/ReflectionMethod.php
loren-osborn: This is when the PHP source code doesn't exist, so we defer to the output of the native implementation.

src/ReflectionParameter.php
loren-osborn: Some built in functions and methods have optional parameters **without** a default value, which provides some interesting introspection values. This is part of the handling of this case.

src/ReflectionParameter.php
loren-osborn: Yes, this is part of the handling for optional parameters without default values, which is new in this PR.

src/ReflectionParameter.php
loren-osborn: Again, this is part of the handling for optional parameters without default values, which is new in this PR.

src/ReflectionParameter.php
loren-osborn: Again, this is part of the handling for optional parameters without default values, which is new in this PR.

src/ReflectionParameter.php
loren-osborn: 👍 Spelling and typing will be the death of me... good catch.

src/ReflectionParameter.php
loren-osborn: Again, this is part of the handling for optional parameters without default values, which is new in this PR. Perhaps I should make `isDefaultValueSet()` private so it's more clearly a internal convienience function, rather than something that's part of the interface.

src/ReflectionProperty.php
loren-osborn: I have to investigate further, but you're probably right. If I can simplify this, I will. 👍

src/Traits/ReflectionFunctionLikeTrait.php
loren-osborn: This is just blatantly wrong. This is an instance method, not a static method.

src/Traits/ReflectionFunctionLikeTrait.php
loren-osborn: As above, it's an instance method.

src/Traits/ReflectionFunctionLikeTrait.php
loren-osborn: I'm thinking similarly. I believe `__call()' is pretty expensive, but I believe this can be simplified. I'll look into it further. :+1:

aik099:
> The goal of this PR is to bring feature pairity to Go\ParserReflection classes with their native equivalents for functions/classes/methods/etc. that are compiled into PHP (and therefore have no parsable PHP source code.)

I thought it was solving #16. Does it as well?

> Of course, we don't get this new flexibility for free, so we have lots of new special cases for most `Go\ParserReflection` classes, for cases when there is no PHP source code. In these cases, parsed node properties, like `$this->functionLikeNode` and `$this->classLikeNode`, are null.

I understand now. Thanks.

src/Traits/ReflectionFunctionLikeTrait.php
loren-osborn: I included tests for two seemingly ubiquitous extensions. I believe I setup the tests to skip if not loaded... not sure I have coverage on this particular method, but I will verify. :+1: :memo:

src/Traits/ReflectionFunctionLikeTrait.php
loren-osborn: If we're in this else block there is no PHP source code to analyze, and we're only querying the PHP binary and any loaded extensions.

src/ReflectionClass.php
```diff
@@ -135,18 +139,4 @@ protected function __initialize()
...
-    /**
-     * Create a ReflectionClass for a given class name.
-     *
-     * @param string $className
-     *     The name of the class to create a reflection for.
-     *
-     * @return ReflectionClass
-     *     The apropriate reflection object.
-     */
-    protected function createReflectionForClass($className)
```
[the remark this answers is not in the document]
loren-osborn: Now that we can reflect builtin classes, this is no longer needed.

src/ReflectionParameter.php
aik099: Better to move this into separate PR for optional parameter without default value handling (part 5).

src/ReflectionParameter.php
aik099: Better to move this into separate PR for optional parameter without default value handling (part 1).

src/ReflectionParameter.php
aik099: Better to move this into separate PR for optional parameter without default value handling (part 2).

src/Traits/ReflectionFunctionLikeTrait.php
aik099: Why this was working before and needs to be changed now? Maybe this was workaround for older PHP versions?

src/Traits/ReflectionFunctionLikeTrait.php
aik099: As above, why this was needed before?

src/ReflectionClass.php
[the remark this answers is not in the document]
aik099: OK.

src/ReflectionFunction.php
loren-osborn: :+1: :memo: (Note to self:) Verify test cases!!!

src/ReflectionParameter.php
loren-osborn: I did some thinking on this. I think I'm able to do the same thing with a more minor change to `ReflectionParameter`. Let me take another crack at this. 

Everything I said above is still valid, but I think I have a more subtle solution with the same behavior.

src/Traits/ReflectionFunctionLikeTrait.php
loren-osborn: I have some thoughts on this I want to explore. No promises, but we'll see.

loren-osborn:
@aik099, I forgot about your asking about #16. The way I read #16, yes it will, even though I generally thought of this PR as solving #17, although indirectly.

src/ReflectionParameter.php
aik099: > optional parameter with no default value

Then what is it? You haven't provided any example PHP code.

> Everywhere else in this PR, if we don't have source code for a function/method/class/etc.

OK. Since this PR doesn't sound like something that can be finalized/merged this year, then I thought at least `optional parameter with no default value` (whatever this is) could be merged as separately PR. I was wrong apparently. No actions needed from you here.

src/ReflectionParameter.php
aik099: OK. See above.

src/ReflectionParameter.php
aik099: OK. See above.

src/ReflectionParameter.php
aik099: OK. See above.

src/ReflectionParameter.php
aik099: OK. See above.

src/Traits/ReflectionFunctionLikeTrait.php
aik099: 👍  for moving this into separate PR.

src/Traits/ReflectionFunctionLikeTrait.php
aik099: 👍  for moving this into separate PR.

src/Traits/ReflectionFunctionLikeTrait.php
aik099: OK.

src/ReflectionParameter.php
loren-osborn: I apologize for not being more responsive lately. I have had far less time to devote to this project than I anticipated, but I am still hopeful to have this merged in, within 2 to 4 weeks with everything else going on.

As far as an example of a builtin function like this, take `preg_match()` as an example:

```php
<?php

$preg_matchRef = new ReflectionFunction('preg_match');
$params = $preg_matchRef->getParameters();
var_export(['getName()' => $params[2]->getName(),
'isOptional()' => $params[2]->isOptional(),
'isDefaultValueAvailable()' => $params[2]->isDefaultValueAvailable()]);
print "\n\n";
```

The `&$matches` parameter (as it appears in the documentation, although it's called `$subpatterns` internally, is optional, but has no default value. To my knowledge, there is no way to represent this in PHP source code.

aik099: > I apologize for not being more responsive lately.

Not to worry. We're not on a schedule here 😉 

> To my knowledge, there is no way to represent this in PHP source code.

Haven't thought about this scenario. I guess this is what you get, when attempt to reflect built-in functions :(

src/Traits/ReflectionFunctionLikeTrait.php
loren-osborn: One other thing, for **this** PR: the native implementation returns `?ReflectionClass`, so the result needs to be reinstantiated as a `Go\ParserReflection\ReflectionClass` before it's returned, unless it's `null`.

aik099: No big deal. Good catch.

loren-osborn: Ok... doesn't appear that the old code has any bad behavior, but I moved this into PR #86.

loren-osborn:
@aik099, continuing the discussion from #86:

Part of the reason I'm not **quite** so scared of it, is this is my second pass at this change, and it gave me no problems previously. I agree it's a bigger delta than I remember. One thing that I probably would have included in this PR, that I think will be it's own PR now, is explicit closure support. (I may add protection **against** using `ReflectionFunction` with a closure to this PR, but merely for the purpose of defensive programming.)

On the subject of a closure support PR, I think it would be useful to add two methods to `ReflectionFunctionLikeTrait`:
* The first for purposes of traversal, (that would have a work-alike method in`ReflectionFileNamespace`) would return an array of all closures defined within the function, method or file's namespace. (It may be a good idea to have aggregator methods in `ReflectionClass` and `ReflectionFile` For this new method.)
* The second I'm a bit less clear on: I'm thinking a static analysis equivalent of `getClosureThis()`. I'm not quite so sure what that would be though. I'm pretty sure it would return `null` whenever `isClosure()` is `false`, and whenever a closure isn't defined within a non-static method, but I'm not sure if it should just return `getClosureScopeClass()` the rest of the time, or something a bit more stateful, without having to instantiate (and load) the class definition.

What do you think? This is only a half-formed idea at this point.

aik099:
> The first for purposes of traversal, (that would have a work-alike method inReflectionFileNamespace) would return an array of all closures defined within the function, method or file's namespace. (It may be a good idea to have aggregator methods in ReflectionClass and ReflectionFile For this new method.)

1. Is PHP built-in reflection capable if this?
2. For what purposes this functionality can be used?

> The second I'm a bit less clear on: I'm thinking a static analysis equivalent of getClosureThis(). I'm not quite so sure what that would be though. I'm pretty sure it would return null whenever isClosure() is false, and whenever a closure isn't defined within a non-static method, but I'm not sure if it should just return getClosureScopeClass() the rest of the time, or something a bit more stateful, without having to instantiate (and load) the class definition.

In real reflection (e.g. created from object) that would probably point back to object to which reflected method belongs. But when you're doing it statistically you have no object to return from that method. The only thing you can return is another reflection of what you expect $this to be, but then return type won't be compatible with what built-in reflection method returns.

loren-osborn:
> > The first for purposes of traversal, (that would have a work-alike method in `ReflectionFileNamespace`) would return an array of all closures defined within the function, method or file's namespace. (It may be a good idea to have aggregator methods in `ReflectionClass` and `ReflectionFile` For this new method.)
>
> 1. Is PHP built-in reflection capable if this?

No, these would be discovery methods like `getFunctions()`, `getClasses()` and `getConstants()` in `ReflectionFileNamespace`.

> 2. For what purposes this functionality can be used?

The intent, is being able to inspect closures that **might** be returned to (or otherwise interact with) calling code. The static analysis to determine **which** closures **may** be returned would be well outside the scope of this module, so the thinking was to simply return an array of all closures defined within the scope being queried. My biggest concern about doing this, is that closures are intrinsically anonymous, so there is no logical way to organize them in a returned array (except order of occurrence, which is likely to change.)

> > The second I'm a bit less clear on: I'm thinking a static analysis equivalent of `getClosureThis()`. I'm not quite so sure what that would be though. I'm pretty sure it would return `null` whenever `isClosure()` is `false`, and whenever a closure isn't defined within a non-static method, but I'm not sure if it should just return `getClosureScopeClass()` the rest of the time, or something a bit more stateful, without having to instantiate (and load) the class definition.
>
> In real reflection (e.g. created from object) that would probably point back to object to which reflected method belongs. But when you're doing it statistically you have no object to return from that method. The only thing you can return is another reflection of what you expect `$this` to be, but then return type won't be compatible with what built-in reflection method returns.

I think I was slightly misunderstood, but mostly because I believe I was being too ambitious:

<ul> <li> First, I think the correct solution here would be adding a method <code>getClosureThisClass()</code>, which would return the same value as <code>getClosureScopeClass()</code>, <b>except</b> that it would return <code>null</code> when the closure is defined within a static method of a class.<ul><li> It would essentially be equivilant to: <div><code><div>(</div><div>&nbsp;&nbsp;&nbsp;&nbsp;($closureReflection-&gt;isClosure() &amp;&amp; $closureReflection-&gt;getClosureThis()) ?</div><div>&nbsp;&nbsp;&nbsp;&nbsp;new ReflectionClass(get_class($closureReflection-&gt;getClosureThis())) :</div><div>&nbsp;&nbsp;&nbsp;&nbsp;null</div><div>)</div></code></div></li> </ul></li> <li> Second, I think you misunderstood what I meant by a "more stateful" value than a <code>ReflectionClass</code> because (in hindsight) what I had in mind was so far beyond the scope of what we would want in this module. I think I could explain best by example: <ul><li> Imagine the following code fragment within a instance method we're introspecting: <div><code><div> if ($this-&gt;getFred() &lt; 7) {</div><div>&nbsp;&nbsp;&nbsp;&nbsp;return (function () { return $this-&gt;getGeorge() + 3; } );</div><div>}</div></code></div></li><li> In this case, a sufficiently advanced (bloated) static analyzer could deduce, if we had a <code>ReflectionFunction</code> of the closure the fragment above returns, some yet-to-be-defined static analog to <code>getClosureThis()</code> could return something that the object it represents would have a method <code>getFred()</code>, that returned a value less than 7. </li> <li> Even ignoring the question of how this state could be presented to the caller, the depth of static analysis required for this would be daunting, and is well beyond what we might want in this composer module.</li> <li> <b>(Summary) This is well outside anything we would want to attempt here.</b></li></ul></li></ul>

aik099:
To summarize better not to implement any of https://github.com/goaop/parser-reflection/pull/82#issuecomment-344711580, because it will be out of scope of this library and  any kind of reflection library shouldn't be doing static analysis of that kind.

loren-osborn:
@aik099, basically, I mostly agree with you, and was really sitting on the fence on most of that **except** I think adding `getClosureThisClass()` is quite narrow in scope, doesn't require any static analysis, and provides a staticly relevant alternative to `getClosureThis()`. (The only reason this method isn't a trivial wrapper over existing functionality, is that it needs to determine the if it was defined within a method of a class, which class, and if it was defined within a static method.) Even seeing the utility of this method, I see it as a low-priority nice-to-have feature, and of limited importance.

src/IReflection.php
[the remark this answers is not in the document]
loren-osborn: @lisachenko, I'm finally digging into making these changes (as time is available), and a concern came to mind: It looks like some of the `ReflectionType` values in PHP 7.0 are being broken out into subclasses in PHP 7.2. If this trend continues, we may see `ReflectionTrait` and `ReflectionInterface` as derived classes of `ReflectionClass` in the not too distant future. If we wish to mirror this class heirarchy, this would cause a name collision for us. Perhaps we should change `IReflection` to `Reflector` which is already in PHP as an interface?

(Just a thought. In the meantime, I've already made the change you suggested. [Not pushed yet.] )

src/ReflectionClass.php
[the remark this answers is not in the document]
loren-osborn: @aik099, I wanted to share the results of verifying all the logical cases were being covered: There was, unfortunately one case that I'm not currently testing. I need to add tests for files that haven't been  included by PHP when the test is performed. (I had already been thinking how to do this properly, but I will come up with a stop-gap measure for now.) Currently we include all the Stub files, so we can compare the results to the behavior of the builtin reflection classes. 

- [ ] Add tests for code not `include`d or `require`d by PHP at test time.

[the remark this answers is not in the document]
loren-osborn: Added cases to the test suite for `Go\ParserReflection\Stub\NeverIncluded` which mirrors `Go\ParserReflection\Stub`, but is never included or required by PHP.

loren-osborn:
My latest changes by no means address all raised concerns yet. I just wanted to share my progress so far. I regret that this PR is taking so long. New in these changes is testing against classes that have **actually** never been loaded by the PHP interpreter. Much need of refactoring all around.

loren-osborn:
Just wanted to note why the build is currently broken:
I uncovered a few typehint names (that are common aliases for PHP builtin types) that PHP actually treats like class names, but the PHP-parser misidentifies them as builtin types. I'm writing a new NodeVisitor to resolve this.

aik099:
> I uncovered a few typehint names (that are common aliases for PHP builtin types) that PHP actually treats like class names, but the PHP-parser misidentifies them as builtin types. I'm writing a new NodeVisitor to resolve this.

Maybe you should report this for PHP-Parser project as well. What if it's fixed in newer PHP-Parser version and we just need to upgrade.

loren-osborn:
I agree it **should** be fixed upstream, but I see no issue shimming it until it’s fixed. I don’t have PHP7 on my dev box, so I’m only seeing it in the Travis build log. I expect Travis is getting the latest PHP-Parser?

The behavior seems a bit odd. The parser is identifying `object` and `integer` as builtin typehints. `integer` should, of course be `int`, and there doesn’t seem to be a typehint for objects. As they aren’t valid typehints, PHP is treating them as class names, and prepending the current namespace to them, as it should.

aik099:
In upstream of https://github.com/nikic/PHP-Parser you mean right? Please check if this wasn't fixed already.

loren-osborn:
This will take me a bit of doing to research, but I just verified that Travis is pulling nikic/php-parser (v3.1.2) and not an unstable master branch, so this may have already been fixed. Will investigate more as limited time permits.

loren-osborn:
@aik099, I did some research on the Php-Parser bug I found and confirmed that:

* It affects the latest development version of the 3.x branch.
* ParserReflection isn't (yet) compatible with the 4.0.0 alpha2 version Php-Parser just released.
* There was no open issue about this problem.

So I opened https://github.com/nikic/PHP-Parser/issues/449 to get them to fix it upstream, but I am working on a temporary fix until they get around to fixing it.

aik099:
Nice, then create NodeVisitor to overcome that problem and mention in it's class description link to associated issue.

loren-osborn:
Exactly!

src/ReflectionFunction.php
loren-osborn: The one case that wasn't being tested was when the function hadn't been loaded by PHP. Added this for both classes and functions.

loren-osborn:
It looks like https://github.com/nikic/PHP-Parser/issues/449 is more an issue of simulating different versions of the PHP parser. Different typehints were added in different PHP versions and `PhpParser` runs either in latest PHP versions of 5 and 7, but no easy way to differentiate PHP 7.1 from PHP 7.2 for instance.

tests/TestCaseBase.php
loren-osborn: @lisachenko, I did some exploring and found `PHPUnit` uses `SebastianBergmann\Exporter\Exporter` to avoid this very mess. I see no reason I can't do similarly. I'm (slowly) working on extracting a PR from this one that refactors the test suite and will address this there.

src/ReflectionInterface.php
```diff
@@ -0,0 +1,15 @@
...
+ * Parser Reflection API
+ *
+ * @copyright Copyright 2015, Lisachenko Alexander <lisachenko.it@gmail.com>
+ *
+ * This source file is subject to the license that is bundled
+ * with this source code in the file LICENSE.
+ */
+namespace Go\ParserReflection;
+
+interface ReflectionInterface
```
loren-osborn: @lisachenko, I'm kind of knee-deep in alligators here, trying to finish and untangle this into two workable PRs here, but I wanted to give you an update relevant to the `Go\ParserReflection\ReflectionInterface` interface I added. I added it believing that `\Reflection` was the parent class of all Reflection classes. I have since discovered (at least in PHP 7) this isn't the case. The `\Reflector` interface, however **is** implemented by all Reflection classes (except, ironically, the purely-static `\Reflection` class). So the new interface `Go\ParserReflection\ReflectionInterface` should clearly be renamed either `Go\ParserReflection\ReflectorInterface` or simply `Go\ParserReflection\Reflector` (as `\Reflector` is already an interface). I welcome your input.

I am continuing work on this PR here: [loren-osborn:losborn_testDirReorgAndPassThrough](https://github.com/loren-osborn/parser-reflection/tree/losborn_testDirReorgAndPassThrough) which will be split into two distinct PRs.

lisachenko:
One year later :)

Guys, could we have clean PR to merge for 2.0?

loren-osborn:
@lisachenko and team, I’m sorry for dragging my feet on this... (It has been a crazy year.) I’ll try to get something cleaned up in the next 2 weeks. (Then again, I’ve been away from this so long i’m almost starting over mentally on this, so any ideas anyone has will likely help me get moving in the right direction.)

icanhazstring:
@loren-osborn don't stress ;)
And even then, mentally starting over here is maybe not even a bad thing. Since things evolve over time. Maybe you find something to improve over your old implementation ;)

lisachenko:
Closing this one as it's 2020 year )

PR #86, opened 2017-11-09
Eliminate nonsencical forward_static_call()s

loren-osborn:
This is an offshoot of #82. This PR eliminates the `forward_static_call()` calls that should not exist in non-static methods. I expanded the test for `setStaticPropertyValue()` and `getStaticPropertyValue()` to try and produce a bad behavior, and was unable to, so the original code **does** seem to work. The problem is that calling `forward_static_call()` from a non-static method doesn't make any sense. My test confirms, at least for the `setStaticPropertyValue()` and `getStaticPropertyValue()` methods, that both the old and new code **works**.

I did not add any tests for `getClosureThis()` and `getClosureScopeClass()`, as I don't think the current implementation can work with closures since `PhpParser\Node\Expr\Closure` doesn't have a `getName()` method, and should fail when either are called on a closure.

@lisachenko,  I mentioned that I found the original commits that added these 4 calls:
* Origin of `ReflectionFunctionLikeTrait` `forward_static_call()`s in methods `getClosureThis()` and `getClosureScopeClass()` commit https://github.com/goaop/parser-reflection/commit/240d5b55ee7b0203a936e28760fb66e9cb7a0ace
* Origin of `ReflectionClassLikeTrait` `forward_static_call()`s in methods `setStaticPropertyValue()`and `getStaticPropertyValue()` commit https://github.com/goaop/parser-reflection/commit/9503a1928796b980bd18c34405da9f4851e1cbc1

Do you have any recollection of why these were added? Was there a specific reason the code was written this way? Am I missing some reason these calls would be apropriate? I welcome your insight.

Thank you.

### DIFF
--- a/src/Traits/ReflectionClassLikeTrait.php
+++ b/src/Traits/ReflectionClassLikeTrait.php
@@ -800,7 +800,7 @@ trait ReflectionClassLikeTrait
     {
         // In runtime static properties can be changed in any time
         if ($this->isInitialized()) {
-            return forward_static_call('parent::getStaticProperties');
+            return parent::getStaticProperties();
         }
 
         $properties = [];
@@ -908,7 +908,7 @@ trait ReflectionClassLikeTrait
     {
         $this->initializeInternalReflection();
 
-        forward_static_call('parent::setStaticPropertyValue', $name, $value);
+        parent::setStaticPropertyValue($name, $value);
     }
 
     private function recursiveCollect(\Closure $collector)
--- a/src/Traits/ReflectionFunctionLikeTrait.php
+++ b/src/Traits/ReflectionFunctionLikeTrait.php
@@ -53,7 +53,7 @@ trait ReflectionFunctionLikeTrait
     {
         $this->initializeInternalReflection();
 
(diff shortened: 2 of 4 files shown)


lisachenko:
Hi! Sorry for late response ) 

Generally, `forward_static_call` functions are not needed indeed. So, you can merge your PR.

loren-osborn:
@lisachenko, while you discussed giving me write access to this repo, I don't think that I've been granted write access yet. Could you or @aik099 merge this? Thanks.

aik099:
Merged, thanks @loren-osborn .

loren-osborn:
@aik099, thank you.

(I'm not intentionally dragging my feet on #82, I just have very scarce resources to devote right now. I hope I can change that soon.)

aik099:
> thank you.

no problem.

> (I'm not intentionally dragging my feet on #82, I just have very scarce resources to devote right now. I hope I can change that soon.)

Sure. I understand that it's very complex PR. Frankly I'm also a bit scared, when reviewing it.

PR #88, opened 2017-11-16
Fix breaking test

VolCh:


aik099:
Have you fixed anything or actually broke something. Right now I see all builds are failing.

VolCh:
Sorry. Test was broken with nikic/php-parser v3.1.1 on my local machine. After update it passes.

PR #89, opened 2017-11-16
Fix problems with leading '\' in FQCNs

### LINKED ISSUE #80: Class \ClassName was not found by locator
The class exists in the top level (no namespaces) in src subdir, it presents in the composer classmap (without leading backslash), the application works fine, but codeception tests with aspectmock does not work:
```
  [InvalidArgumentException]               
  Class \ClassName was not found by locator  
                                           

Exception trace:
 () at /app/vendor/goaop/parser-reflection/src/ReflectionEngine.php:125
 Go\ParserReflection\ReflectionEngine::locateClassFile() at /app/vendor/goaop/parser-reflection/src/ReflectionEngine.php:140
 Go\ParserReflection\ReflectionEngine::parseClass() at /app/vendor/goaop/parser-reflection/src/ReflectionClass.php:44
...
```
PHP versions 5.6 and 7.1, parser-reflection 1.4.0

VolCh:
I found a inconsistent behavior about classname in native \Reflection* and \Go\ParserReflection\Reflection* 

Is it unexpected behavior? The native Reflection* constructors drop the leading slash in the string class name parameter. I made changes in constructors to consist behavior and pass thees tests.  It's ready to push. Also it closes #80.

### DIFF
--- a/src/ReflectionMethod.php
+++ b/src/ReflectionMethod.php
@@ -52,7 +52,7 @@ class ReflectionMethod extends BaseReflectionMethod
         ReflectionClass $declaringClass = null
     ) {
         //for some reason, ReflectionMethod->getNamespaceName in php always returns '', so we shouldn't use it too
-        $this->className        = $className;
+        $this->className        = ltrim($className, '\\');
         $this->declaringClass   = $declaringClass;
         $this->functionLikeNode = $classMethodNode ?: ReflectionEngine::parseClassMethod($className, $methodName);
 
--- a/src/ReflectionProperty.php
+++ b/src/ReflectionProperty.php
@@ -60,7 +60,7 @@ class ReflectionProperty extends BaseReflectionProperty
         Property $propertyType = null,
         PropertyProperty $propertyNode = null
     ) {
-        $this->className    = $className;
+        $this->className = ltrim($className, '\\');
         if (!$propertyType || !$propertyNode) {
             list ($propertyType, $propertyNode) = ReflectionEngine::parseClassProperty($className, $propertyName);
         }
--- a/src/Locator/CallableLocator.php
+++ b/src/Locator/CallableLocator.php
@@ -36,6 +36,6 @@ class CallableLocator implements LocatorInterface
      */
     public function locateClass($className)
     {
-        return call_user_func($this->callable, $className);
+        return call_user_func($this->callable, ltrim($className, '\\'));
     }
 }
--- a/src/Locator/ComposerLocator.php
+++ b/src/Locator/ComposerLocator.php
@@ -51,7 +51,7 @@ class ComposerLocator implements LocatorInterface
      */
     public function locateClass($className)
     {
-        $filePath = $this->loader->findFile($className);
+        $filePath = $this->loader->findFile(ltrim($className, '\\'));
         if (!empty($filePath)) {
             $filePath = PathResolver::realpath($filePath);
         }
--- a/src/LocatorInterface.php
+++ b/src/LocatorInterface.php
@@ -19,7 +19,7 @@ interface LocatorInterface
     /**
      * Returns a path to the file for given class name
      *
-     * @param string $className Name of the class
+     * @param string $className Name of the class (with or without leading '\' FQCN)
      *
      * @return string|false Path to the file with given class or false if not found
      */
--- a/src/ReflectionClass.php
+++ b/src/ReflectionClass.php
@@ -33,7 +33,7 @@ class ReflectionClass extends InternalReflectionClass
      */
     public function __construct($argument, ClassLike $classLikeNode = null)
     {
-        $fullClassName       = is_object($argument) ? get_class($argument) : $argument;
+        $fullClassName       = is_object($argument) ? get_class($argument) : ltrim($argument, '\\');
         $namespaceParts      = explode('\\', $fullClassName);
         $this->className     = array_pop($namespaceParts);
         // Let's unset original read-only property to have a control over it via __get
--- a/tests/Locator/CallableLocatorTest.php
+++ b/tests/Locator/CallableLocatorTest.php
@@ -6,11 +6,12 @@ class CallableLocatorTest extends \PHPUnit_Framework_TestCase
     public function testLocateClass()
     {
         $callable = function ($class) {
-            return $class . '.php';
+            return ltrim($class, '\\') . '.php';
         };
 
         $locator = new CallableLocator($callable);
 
         $this->assertSame('class.php', $locator->locateClass('class'));
+        $this->assertSame('class.php', $locator->locateClass('\class'));
     }
 }
--- a/tests/Locator/ComposerLocatorTest.php
+++ b/tests/Locator/ComposerLocatorTest.php
@@ -13,5 +13,9 @@ class ComposerLocatorTest extends \PHPUnit_Framework_TestCase
             $reflectionClass->getFileName(),
             $locator->locateClass(ReflectionClass::class)
         );
+        $this->assertSame(
+            $reflectionClass->getFileName(),
+            $locator->locateClass('\\' . ReflectionClass::class)
+        );
     }
 }
--- a/tests/Stub/Issue44/Locator.php
+++ b/tests/Stub/Issue44/Locator.php
@@ -11,7 +11,7 @@ class Locator implements LocatorInterface
      */
     public function locateClass($className)
     {
-        if ($className === '\\Stub\\Issue44\\ClassWithNamespace') {
+        if ($className === ClassWithNamespace::class) {
             return __DIR__ . '/ClassWithNamespace.php';
         }
 
--- a/tests/Stub/Issue44/Locator.php
+++ b/tests/Stub/Issue44/Locator.php
@@ -11,7 +11,7 @@ class Locator implements LocatorInterface
      */
     public function locateClass($className)
     {
-        if ($className === ClassWithNamespace::class) {
+        if (ltrim($className, '\\') === ClassWithNamespace::class) {
             return __DIR__ . '/ClassWithNamespace.php';
         }
 


tests/ReflectionClassTest.php
```diff
@@ -290,4 +290,37 @@ protected function getGettersToCheck()
 
         return $allNameGetters;
     }
+
+    /**
+     * Tests getName() method
+     *
+     * @dataProvider getClassNamesToCheck
```
aik099: If I'm not mistaken, then `@dataProvider` annotation should go after everything else. Or do as done in other places.

tests/ReflectionClassTest.php
```diff
@@ -290,4 +290,37 @@ protected function getGettersToCheck()
...
+     *
+     * @param string $className
+     */
+    public function testClassName($className) {
```
aik099: The `{` should be on next line as in other places.

tests/ReflectionClassTest.php
```diff
@@ -290,4 +290,37 @@ protected function getGettersToCheck()
...
+
+        $this->assertSame(
```
aik099: Interesting. I the `getName` method result is compared in `testReflectionMethodParity` test, but one of 2 things happens:

1. either `->getName()` result is correct AND `->name` isn't correct
2. or class names list used for testing isn't complete

Likely you should add classes from global namespace to these tests as well.

loren-osborn: Should we add FQ names for classes and functions as distinct cases in the `@dataProvider`s for `ReflectionClassTest` and `ReflectionFunctionTest`. (We should also similarly add case-insensitivity test cases to the `@dataProvider`s for `ReflectionFunctionTest` and `ReflectionMethodTest` While we're at it.)

tests/ReflectionClassTest.php
VolCh: Technically it doesn't matter where.  II use style of ReflectionClassTest::testGetModifiers() & ReflectionClassTest::testReflectionMethodParity()

VolCh:
I implemented testReflectionPropertyParity() method (with a refactor of testReflectionMethodParity()) to check parity of property. All tests are passed. Then I added FQCN to all parity checks. 

Tests are broken now with 100+ failures!

aik099:
I'm not sure if these are good news or bad news.

Are tests failing due false positives or they really indicate that something not working consistently?

VolCh:
My last commit emulate original PHP behavior. I explored PHP source about class name related behavior functions & methods: all of them drop leading slash first of all, before parsing or search or save. So I think this is not false positive result.

VolCh:
@aik099 what do you think about it?

src/ReflectionMethod.php
```diff
@@ -52,7 +52,8 @@ public function __construct(
         ReflectionClass $declaringClass = null
     ) {
         //for some reason, ReflectionMethod->getNamespaceName in php always returns '', so we shouldn't use it too
-        $this->className        = $className;
+        $this->className        = is_object($className) ? get_class($className) : ltrim($className, '\\');
```
aik099: Does this class even support passing an object in `$className` parameter? I'm suspecting it doesn't, because parameter name suggests that. Yeah, I know that PHP built-in reflection supports that.

If it doesn't, then do `ltrim` all the time.

src/ReflectionProperty.php
```diff
@@ -60,7 +60,8 @@ public function __construct(
         Property $propertyType = null,
         PropertyProperty $propertyNode = null
     ) {
-        $this->className    = $className;
+        $this->className    = is_object($className) ? get_class($className) : ltrim($className, '\\');
```
aik099: Does this class even support passing an object in `$className` parameter? I'm suspecting it doesn't, because parameter name suggests that. Yeah, I know that PHP built-in reflection supports that.

If it doesn't, then do `ltrim` all the time.

tests/ReflectionClassTest.php
```diff
@@ -47,21 +48,21 @@ public function testGetModifiers($fileName)
...
+        \ReflectionClass $refClass,
         ReflectionClass $parsedClass,
         $getterName
     ) {
         $className = $parsedClass->getName();
-        $refClass  = new \ReflectionClass($className);
 
-        $expectedValue = $refClass->$getterName();
-        $actualValue   = $parsedClass->$getterName();
+        $expectedValue = $refClass->{$getterName}();
```
aik099: Is this and next change really needed? 

Should work fine on both PHP5 and PHP7 already.

tests/ReflectionClassTest.php
```diff
@@ -47,21 +48,21 @@ public function testGetModifiers($fileName)
...
      *
-     * @param ReflectionClass   $parsedClass Parsed class
-     * @param \ReflectionMethod $refMethod Method to analyze
-     * @param string                  $getterName Name of the reflection method to test
+     * @param \ReflectionClass $refClass Original reflection class
+     * @param ReflectionClass $parsedClass Parsed class
+     * @param string $getterName Name of the reflection method to test
      */
     public function testReflectionMethodParity(
+        \ReflectionClass $refClass,
```
aik099: I guess `methodCaseProvider` now returns more data, but this particular test doesn't use that extra data?

tests/ReflectionClassTest.php
```diff
@@ -290,4 +300,58 @@ protected function getGettersToCheck()
...
+        ];
+
+        return $names;
+    }
+
+    /**
+     * @param string[] $membersToCheck
+     * @return array
+     */
+    protected function caseProvider($membersToCheck)
```
aik099: I see you've moved method down, but because of that it's hard to see what you've changed in the method itself.

tests/ReflectionClassTest.php
```diff
@@ -290,4 +300,58 @@ protected function getGettersToCheck()
...
+                include_once $fileName;
+                foreach ($reflectionFile->getFileNamespaces() as $fileNamespace) {
+                    foreach ($fileNamespace->getClasses() as $refClass) {
+                        $qcn = ltrim($refClass->getName(), '\\'); // workaround for #80
+                        $fqcn = '\\' . $qcn;
+                        $classes = [
+                            $qcn => [$refClass, new \ReflectionClass($qcn)],
+                            $fqcn => [new ReflectionClass($fqcn), new \ReflectionClass($fqcn)],
+                        ];
+                        foreach ($classes as $caseName => list($parsedClass, $originalClass)) {
```
aik099: 1. Is this `list(` function usage compatible with all PHP versions where library itself can be used?
2. Please rename `$originalClass` into `$builtinClass`. Maybe it makes sense to rename corresponding parameter in method, where this data provider is used (e.g. `$refMethod` to `$builtinMethod` and such).

tests/ReflectionClassTest.php
```diff
@@ -290,4 +300,58 @@ protected function getGettersToCheck()
...
+        foreach ($files as $fileList) {
+            foreach ($fileList as $fileName) {
+                $fileName = stream_resolve_include_path($fileName);
+                $fileNode = ReflectionEngine::parseFile($fileName);
+
+                $reflectionFile = new ReflectionFile($fileName, $fileNode);
+                include_once $fileName;
+                foreach ($reflectionFile->getFileNamespaces() as $fileNamespace) {
+                    foreach ($fileNamespace->getClasses() as $refClass) {
+                        $qcn = ltrim($refClass->getName(), '\\'); // workaround for #80
```
aik099: This PR should be ltrimming class name. Why we need to also ltrim in tests?

VolCh: Tests should check that no difference with and without leading backslash between native Reflection* and Go\ParserReflection\Reflection*. It is common consists checking regardless fix in the PR.

tests/ReflectionClassTest.php
VolCh: 1. I used list() in PHP3 era :) Changes in 7.0 does not affect this simple use case.
2. ok

src/ReflectionMethod.php
VolCh: \Go\ParserReflection\ReflectionClass supports (documented like string|object) and is_object check present. I decide 'casting' object to string in \Go\ParserReflection\ReflectionMethod & \Go\ParserReflection\ReflectionProperty make them more consist with built-in & \Go\ParserReflection\ReflectionClass and provide expected behavior. Do you agree?

tests/ReflectionClassTest.php
aik099: 1 - you could use `list(`, but support for using `list` right within `foreach` construct was added later

tests/ReflectionClassTest.php
aik099: I'm still not getting it. If this PR normalizes stuff so that there won't be any leading `\` in class names around the library and built-in reflection already does it, then why we need to ltrim it in tests?

src/ReflectionMethod.php
aik099: I understand now. Since library only supports class names and not objects you've created workaround code to convert object (if somebody gives it by mistake) into a class so that library won't have any issues.

Nice catch, but this (and associated place in `ReflectionProperty` and other places) needs to be moved into separate PR with corresponding tests. I'd better not to mix this with topic of this PR.

tests/ReflectionClassTest.php
loren-osborn: I just checked this, and it appears to work in [PHP 5.5.0 and above](https://3v4l.org/37Ok9). (I don’t think we support anything older than 5.6.0, so I suspect this will be ok. @aik099, can you confirm this?)

aik099: What I've meant is for @VolCh to find http://php.net/manual/en/control-structures.foreach.php#control-structures.foreach.list, which clearly says that it works since PHP 5.5. That doesn't quite qualify for `I used list() in PHP3 era` :)

Yeah, @loren-osborn , it works according to docs. I haven't used such syntax myself though.

loren-osborn:
Not that it particularly matters, but the changes in #82 to `tests/` is going to be pretty frustrating to merge with this, as i’ve broken out many of the loops into separate tests. Maybe you want to try to parallel some of that to ease the merge? 😕

...or not.

aik099:
Maybe you can do tests refactoring in separate PR, that can be quickly merged? By the way both PRs are made by you so you're blaming yourself here 😉

loren-osborn:
This isn’t my PR. I don’t know what you mean about blaming myself here.

I **HAD** started a test suite refactor branch [here: loren-osborn/parser-reflection:losborn_testDirReorg_orig](https://github.com/loren-osborn/parser-reflection/commits/losborn_testDirReorg_orig) but needed to focus my attention in other direction at the time for various reasons. It was a pretty big change and will be a very hairy merge at this point, but I have borrowed code from it in the last couple PRs I’ve submitted.

I know i’ve had trouble making any practical headway on #82 in my meager 1-2 hours a week on my XP machine that barely runs PHP 5.6, so I’m concerned that breaking off another PR of the complexity of [losborn_testDirReorg_orig](https://github.com/loren-osborn/parser-reflection/tree/losborn_testDirReorg_orig) may seriously derail #82.

aik099:
Now I see, you've (@loren-osborn) mentioned on this PR (authored by @VolCh) that there might be issues for merge.

loren-osborn:
@aik099, I did a mental recap of the branch I mentioned above, and realized I had been trying to do too much in one PR, and got stuck on what **should** have been the third of four distinct PRs.

What I had intended for this branch:

1. Reorganize the test directory, separating test classes from stub and support files.
    * The main reason for this was to put support code in a separate area, and make a logical place in the test class directory to test the test-support code.
    * This was in anticipation of refactoring the “meat” of most of the `@dataProvider`s of comparators used for asserts into support classes, keeping the complexity **out** of the test classes.
2. Move the test code into appropriate namespaces.
3. Get to 100% code coverage.
    * This is a big job by itself.
    * At least one of the classes I wanted to test was best done by mocking built-in functions, so I had started work on a bare bones, ApectMock like, tool to do this in a limited capacity.
4. Refactor the **existing** test code. 

I think 1 & 2 naturally go together.  Given the branch is hopelessly out of date, I think the best strategy is to merge it into #82, separate out 1 & 2 together, and 4 by itself into new PRs, pull out 3 to live on it’s own, longer term, branch, and leave the rest behind in #82. This will allow me to leverage the test suite changes I already made in #82.

aik099:
Do whatever you see fit. From PR reviewing point the smaller and on topic PR is the better to me.

> Move the test code into appropriate namespaces.

Aren't test case classes already namespaced?

loren-osborn:
@aik099, they share a namespace with the production code. In my opinion this isn’t really appropriate.

aik099:
> ... they share a namespace with the production code. In my opinion this isn’t really appropriate, especially because they don’t live in the production code directories. This also prevents the composer auto loader from working with these files.

There are 2 techniques here:

* like in Symfony and here, where there is `Tests` sub-namespaces within application namespace (e.g. `Vendor\AppName\ClassName` + `Vendor\AppName\Tests\ClassName`)
* like you say (and I like to do) to have `Tests` sub-namespace above application namespace (e.g. `Vendor\AppName\ClassName` + `Tests\Vendor\AppName\ClassName`)

Plus for having `Tests\` in front of namespaces is that you can move auto-loading into `autoload-dev` in `composer.json` and then optimized auto-loader in production environment won't have code for auto-completing test classes.

loren-osborn:
@aik099, we’re talking about stuff I did a few months ago, and haven’t really touched since. If I recall currently, I believe I used the first strategy you mentioned above. I also think I was mistaken about composer not working with a namespace spread across multiple directories.

loren-osborn:
@aik099, I definitely still have a great deal of refactoring and untangling to do before issuing a PR for this, but I wanted to show you that I have gotten these changes merged into the current code base, and have all tests passing again. Here is my current working branch for this: https://github.com/loren-osborn/parser-reflection/tree/losborn_testDirReorgAndPassThrough

VolCh:
@aik099 sorry for the silence. I have been removed object casting to string in ReflectionMethod & Property.

But I don't know what to do with tests. Are so big changes acceptable now? I can make just three small tests to check backslash presence/absence behavior.

aik099:
At this point I'm really thinking, that purpose of the fix (considering amount of changes made) isn't worth it. People were using library with this problem before and haven't complained that much :(

Considering speed of this PR development I need to do full review on each push and that takes a lot of time.

lisachenko:
It’s pity, but I have a feeling that we should abandon this project and switch to BetterReflection, because it’s well maintained, whereas we can’t even merge this old PR with huge amount of changes :( Probably, I should work hard on this, but I’ lack of time....

Guys, what do you think? Should we keep working on this library?

aik099:
I'd keep it in current state with policy for accepting small features/bug fixes. It does it job well currently and have wider PHP version support then https://github.com/Roave/BetterReflection.

There are workarounds for known issues, but fixing these issues in the library itself would be too time consuming.

In fact in some reflection-related applications it's better to create thin wrapper around https://github.com/nikic/PHP-Parser instead of using full-featured reflection provided by this or better-reflection library.

VolCh:
I really did not think that my PRs would raise the question about abandoning the library. :(

In this case there is no point in globally changing the tests. I have done a minor changes in the locators and class related Reflection* constructors to fix problems with or without leading backslash in any cases. 

If someone use locators directly this PR shouldn't break the code.

I can remove some changes if they are not acceptable by the policy for accepting small features/bug fixes. Also I save previous changes in https://github.com/VolCh/parser-reflection/tree/rewrite-tests

loren-osborn:
@aik099, @lisachenko, I've continued devoting what scarce resources I've had to PRs for this project, and am optimistic we may have, at least something to offer merging with BetterReflection, (for instance, we still support PHP 5.5 and 5.6). I hate to admit the time I've invested in recent months may have been for naught, but it's silly "throwing good money after bad".

Has anyone investigated what we have to offer that BetterReflection doesn't do? I'd be eager to hear what you find.

If it's relevant, I've continued to make steady, but glacial progress on PRs for this project.

aik099:
Support for PHP 5.x version at least.

aik099:
@VolCh , is this PR ready for merging?

VolCh:
@aik099 Yes, it is ready.

aik099:
Merging, thanks @VolCh .

00eadeyemi:
@aik099 do you have any idea when the next release is going to be? I am not sure setting my dependency version to dev-master is a good idea but yet I really need this fix.

aik099:
No idea.

lisachenko:
@00eadeyemi patch version 1.4.1

00eadeyemi:
@lisachenko Thank you !!!

PR #92, opened 2018-11-25
Update version of nikic/php-parser

### LINKED ISSUE #91: Bump  "nikic/php-parser": "^1.2|^2.0|^3.0", add ^4.0, if possible
SF 4 (web app) installs maker bundle, which requires nikic/php-parser ^4.0, and then we have a conflict...

If nikic/php-parser can be bumped to ^4.0 as well, all could be just fine.

Is it possible to do it safely ?

JustBlackBird:
Fixes #91

### DIFF
--- a/src/NodeVisitor/RootNamespaceNormalizer.php
+++ b/src/NodeVisitor/RootNamespaceNormalizer.php
@@ -44,7 +44,7 @@ class RootNamespaceNormalizer extends NodeVisitorAbstract
             }
         }
         // Wrap all statements into the namespace block
-        $globalNamespaceNode = new Namespace_(new FullyQualified(''), array_slice($nodes, $lastDeclareOffset));
+        $globalNamespaceNode = new Namespace_(null, array_slice($nodes, $lastDeclareOffset));
         // Replace top-level nodes with namespaced node
         array_splice($nodes, $lastDeclareOffset, count($nodes), [$globalNamespaceNode]);
 
--- a/src/NodeVisitor/StaticVariablesCollector.php
+++ b/src/NodeVisitor/StaticVariablesCollector.php
@@ -61,7 +61,13 @@ class StaticVariablesCollector extends NodeVisitorAbstract
                     $value = null;
                 }
 
-                $this->staticVariables[$staticVariable->name] = $value;
+                if ($staticVariable->var->name instanceof Node\Expr) {
+                    $expressionSolver->process($staticVariable->var->name);
+                    $name = $expressionSolver->getValue();
+                } else {
+                    $name = $staticVariable->var->name;
+                }
+                $this->staticVariables[$name] = $value;
             }
         }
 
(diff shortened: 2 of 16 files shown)


lisachenko:
Thank you for this PR! This should be next major

src/NodeVisitor/RootNamespaceNormalizer.php
```diff
@@ -44,7 +44,7 @@ public function beforeTraverse(array $nodes)
             }
         }
         // Wrap all statements into the namespace block
-        $globalNamespaceNode = new Namespace_(new FullyQualified(''), array_slice($nodes, $lastDeclareOffset));
+        $globalNamespaceNode = new Namespace_(null, array_slice($nodes, $lastDeclareOffset));
```
lisachenko: Is it ok to use `null` here instead of FQN? This may require some checks from outside...

src/NodeVisitor/StaticVariablesCollector.php
```diff
@@ -61,7 +61,13 @@ public function enterNode(Node $node)
                     $value = null;
                 }
 
-                $this->staticVariables[$staticVariable->name] = $value;
+                if ($staticVariable->var->name instanceof Node\Expr) {
```
lisachenko: Just curious when name for static variable can be an expression?

JustBlackBird: [StaticVar](https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/Stmt/StaticVar.php) is now a wrapper for a [Variable](https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/Expr/Variable.php). The last one in turn can be defined dynamically (`$a = 'foo'; $$a = 'bar';`). Frankly, I don't know any cases when such a dynamic nature can be applied to static vars so yeah, it's a kind of overengenering here :)

src/NodeVisitor/RootNamespaceNormalizer.php
JustBlackBird: By the way: there is not big deal for clients 'case name of [ReflectionFileNamespace](https://github.com/goaop/parser-reflection/blob/master/src/ReflectionFileNamespace.php) is accessible only with [ReflectionFileNamespace::getName()](https://github.com/goaop/parser-reflection/blob/master/src/ReflectionFileNamespace.php#L245) method with such contents:

```php
    public function getName()
    {
        $nameNode = $this->namespaceNode->name;
        return $nameNode ? $nameNode->toString() : '';
    }
```

JustBlackBird: Added a test case to validate behavior of `ReflectionFileNamespace::getName()` is still the same.

lisachenko:
@JustBlackBird Merged, thanks! Sorry for the delay..

PR #96, opened 2019-11-27
getNamespaceAliases() behavior changed in 2.0

MaximilianKresse:
## What to do?
```php
// file to parse
use SomeClass\WithoutAlias;
use AnotherClass\WithAlias as SomeSuperClass;


// parse the file somewhere else
$result = $reflectionNamespace->getNamespaceAliases();

```

## Expected result
```
$result = [
    'SomeClass\WithoutAlias' => 'WithoutAlias'
    'AnotherClass\WithAlias' => 'SomeSuperClass'
];
```

## Actual result
```
$result = [
    'SomeClass\WithoutAlias' => null
    'AnotherClass\WithAlias' => 'SomeSuperClass'
];
```

## Explanation
The result of getNamespaceAliases() was changed in version 2.0 through a change in php-parser v4.

In php-parser v3 you can find this fallback logic for "use"-statements without an "as" (so without an alias): https://github.com/nikic/PHP-Parser/blob/3.x/lib/PhpParser/Builder/Use_.php#L53

The logic was moved in v4 from the builder: https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Builder/Use_.php#L44

To the node: https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/Stmt/UseUse.php#L46

In the parser-reflection you only use the public property "alias" of the node and not the "getAlias" method: https://github.com/goaop/parser-reflection/blob/master/src/ReflectionFileNamespace.php#L457

This pull request fixes this behavior and will result in the same result like in version 1.

I also added unit tests for this because there was none without an alias.

### DIFF
--- a/src/ReflectionFileNamespace.php
+++ b/src/ReflectionFileNamespace.php
@@ -454,7 +454,7 @@ class ReflectionFileNamespace
                 $useAliases = $namespaceLevelNode->uses;
                 if (!empty($useAliases)) {
                     foreach ($useAliases as $useNode) {
-                        $namespaceAliases[$useNode->name->toString()] = $useNode->alias;
+                        $namespaceAliases[$useNode->name->toString()] = (string) $useNode->getAlias();
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


aik099:
Won't be there any BC break, when v3 PHP-Parser is used?

MaximilianKresse:
?

This fix is only for the actual master/version which requires PHP-Parser ^4.0.
https://github.com/goaop/parser-reflection/blob/master/composer.json#L26

lisachenko:
Let's proceed with 2.1.0

lisachenko:
Thanks!

PR #98, opened 2019-12-04
PHP7.4 compatibility fix.

ddinchev:
As discussed here:
https://github.com/goaop/framework/issues/434

As this doesn't seem to cause any issues with versions pre 7.4 - can we merge it under bugfix version to unblock projects migrating to 7.4?


lisachenko:
Looks good for me: https://3v4l.org/vRnFP

lisachenko:
Thank you, will be published as 2.1.0 soon

PR #106, opened 2020-09-22
Added support for nikic/php-parser > 4.6

xfuturomax:


lisachenko:
Hello! Thank you for this PR! I think it should be next major version to avoid installation of php-parser > 4.6 again to existing projects.

src/ReflectionEngine.php
```diff
@@ -281,4 +281,8 @@ public static function parseFileNamespace($fileName, $namespaceName)
         throw new ReflectionException("Namespace $namespaceName was not found in the file $fileName");
     }
 
+    public static function getLexer()
```
lisachenko, line 284: Please add a return type declaration for this method

lisachenko (changes requested):

src/ReflectionEngine.php
xfuturomax: Done

lisachenko (approved):

alekc:
Hi @lisachenko , do you have an approximate frame for the next version (as in when it's going to hit mainstream without having to use hacks in composer json)? 
Cheers

PR #108, opened 2020-11-12
Improve PHPUnit namespace and some stuffs

peter279k:
# Changed log

- Using the `PHPUnit` `^7.0` version to support the `php-7.1+` versions.
- Using the `PHPUnit\Framework\TestCase` to support `PHPUnit ^7` version.
- The PHPUnit `^7` version about fixture is `protected function setUp(): void`.
- Since `php-7.4` version is released, the `implode` function arguments are `implode($glue, $pieces)` . It should be swapped.

The deprecated message is follows:
`PHP Deprecated:  implode(): Passing glue string after array is deprecated. Swap the parameters`

The more details about `implode` is available on [official PHP doc](https://www.php.net/manual/en/function.implode.php#refsect1-function.implode-changelog).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     "nikic/php-parser": "^4.0"
   },
   "require-dev": {
-    "phpunit/phpunit": "~4.0"
+    "phpunit/phpunit": "^7.0"
   },
   "extra": {
     "branch-alias": {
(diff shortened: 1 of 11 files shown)


lisachenko (approved):

lisachenko:
Merged, thanks!

PR #122, opened 2022-11-23
PHP 8 Support

WalterWoshid:
See https://github.com/goaop/framework/pull/486


src/Traits/ReflectionClassLikeTrait.php
```diff
@@ -401,46 +425,19 @@ function (array &$result, \ReflectionClass $instance, $isParent) {
...
-
-        if ($this->isInterface()) {
-            $abstractMethods = $this->getMethods();
-        } else {
-            $abstractMethods = $this->getMethods(\ReflectionMethod::IS_ABSTRACT);
-        }
-        if (!empty($abstractMethods)) {
-            $modifiers += \ReflectionClass::IS_IMPLICIT_ABSTRACT;
-        }
+        $this->initializeInternalReflection();
```
lisachenko, line 432: It is pity that this method triggers now loading of internal core reflection, and AFAIK, this method was used somewhere in code-generation part too. This means it can be break logic of weaving data inside class (we use static analysis, because we run **before** composer autoloader, thus class can not be loaded here properly and can not be reflected)

src/Traits/ReflectionClassLikeTrait.php
```diff
@@ -26,13 +26,18 @@
...
+use ReflectionMethod as BaseReflectionMethod;
+use ReflectionObject as BaseReflectionObject;
+use ReflectionProperty as BaseReflectionProperty;
 
 /**
  * General class-like reflection
+ *
+ * @template T of object
  */
 trait ReflectionClassLikeTrait
```
lisachenko, line 42: I dont't see any support for PHP8 attributes via AST reflection - thus, it won't work properly inside `goaop` itself, as main requirement for `parser-relfection` is to avoid loading class into memory before it will be transformed by the engine.

Is it possible to add support for PHP8 attributes via AST analysis for all reflection classes?

src/ReflectionClassConstant.php
```diff
@@ -164,33 +165,41 @@ public function getValue()
...
         return $this->classConstantNode->isPublic();
     }
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
+     */
+    public function getAttributes(?string $name = null, int $flags = 0): array
+    {
+        return parent::getAttributes($name, $flags);
```
lisachenko, line 196: Unfortunately, this violates main principle of library - to avoid loading of class during analysis. 

This part requires some code, which **uses only AST information from node** to emulate original behavior. Should be relatively easy to implement starting from `nikic/php-parser >= 4.10.0`, see https://github.com/nikic/PHP-Parser/releases/tag/v4.10.0

src/ReflectionExtension.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+namespace Go\ParserReflection;
+
+use ReflectionExtension as BaseExtension;
+
+/**
+ * General parser-level reflection extension
+ */
+class ReflectionExtension extends BaseExtension
```
lisachenko, line 27: We can avoid this class, it is impossible (without `lisachenko/z-engine` of course) to create extension class from raw PHP source code and just from an AST tree :)

src/ReflectionMethod.php
```diff
@@ -100,23 +103,25 @@ public function __toString()
...
+        /** @noinspection PhpFormatFunctionParametersMismatchInspection */
         return sprintf(
-            "%sMethod [ <user%s%s%s>%s%s%s %s method %s ] {\n  @@ %s %d - %d{$paramFormat}{$returnFormat}\n}\n",
+            "%sMethod [ <user%s%s%s>%s%s%s %s method %s ] {\n  @@ %s %d - %d$paramFormat$returnFormat\n}\n",
             $this->getDocComment() ? $this->getDocComment() . "\n" : '',
-            $prototype ? ", overwrites {$prototypeClass}, prototype {$prototypeClass}" : '',
+            $prototype ? ", overwrites $prototypeClass, prototype $prototypeClass" : '',
             $this->isConstructor() ? ', ctor' : '',
-            $this->isDestructor() ? ', dtor' : '',
+            // $this->isDestructor() ? ', dtor' : '',
```
lisachenko, line 123: This line can be removed at all, as PHP8 doesn't report this `dtor` suffix anymore https://3v4l.org/G1hcX

src/ReflectionParameter.php
```diff
@@ -207,13 +214,14 @@ public function getClass()
                     return $this->getDeclaringClass()->getParentClass();
                 }
 
-                throw new ReflectionException("Can not resolve a class name for parameter");
+                return null;
             }
             $className = $parameterType->toString();
 
             $classOrInterfaceExists = class_exists($className, false) || interface_exists($className, false);
```
lisachenko, line 221: For PHP8.1 there should be `enum_exists` call + check everywhere to support enums as well.

src/ReflectionProperty.php
```diff
@@ -106,33 +107,18 @@ public function __debugInfo(): array
         ];
     }
 
-    /**
-     * Return string representation of this little old property.
-     *
-     * @return string
-     */
-    public function __toString()
```
lisachenko, line 114: Why you have decided to remove this method? Now it requires loading of internal metadata of reflection, which is against idea of this library to provide information without loading core data ((

src/ReflectionProperty.php
```diff
@@ -190,43 +176,53 @@ public function getValue($object = null)
...
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
      */
-    public function isDefault()
+    public function isDefault(): bool
     {
-        // TRUE if the property was declared at compile-time
+        $this->initializeInternalReflection();
```
lisachenko, line 183: See my general comment - it should not be allowed for this library to load core reflection, as `goaop/framework` operates **before autoloader** and this is why if you will try to initialize reflection inside weaving process, you may end up with "class not exists" error :(

lisachenko: So, please keep as much methods without internal reflection initialization

src/ReflectionProperty.php
```diff
@@ -190,43 +176,53 @@ public function getValue($object = null)
...
+        return parent::isDefault();
     }
 
     /**
      * {@inheritDoc}
      */
-    public function isPrivate()
+    public function getDefaultValue(): mixed
+    {
+        $this->initializeInternalReflection();
```
lisachenko, line 193: Sorry, this won't work properly for AOP needs because of autoloader process restrictions, `goaop` **must not** use core reflection all the time, otherwise AOP won't work at all ((

lisachenko (changes requested):
I really appreciate this huuuge work ) A lot of work done, but still there are few places which are missing:

1. Attributes support via AST analysis - reflection should not use original core classes and instead AST data must be used to emulate original logic.
2. In some places, I have noticed that you switched original logic to call internal parent implementation - this will trigger autoloading process and can break desired requirement to operate fully on AST data, without loading items in PHP engine. Because once loaded into PHP memory, such items won't be processed by AOP engine

WalterWoshid:
> I really appreciate this huuuge work ) A lot of work done, but still there are few places which are missing:
> 
> 1. Attributes support via AST analysis - reflection should not use original core classes and instead AST data must be used to emulate original logic.
> 2. In some places, I have noticed that you switched original logic to call internal parent implementation - this will trigger autoloading process and can break desired requirement to operate fully on AST data, without loading items in PHP engine. Because once loaded into PHP memory, such items won't be processed by AOP engine

Oh, I understand now, thank you for the information and sorry for not understanding what the purpose of this library was

WalterWoshid:
I see your comments and I don't understand **all** of it ^^" sorry for the complications

lisachenko:
@WalterWoshid you shouldn't excuse at all! You have done really awesome job )) 

I guess this is because of my comments style - I should try write them in a more friendly way, but it is much easier just to comment tech code, so excuse me too

WalterWoshid:
> I should try write them in a more friendly way, but it is much easier just to comment tech code

I love comments 👀

src/ReflectionClass.php
```diff
@@ -112,9 +134,7 @@ public static function collectTraitsFromClassNode(ClassLike $classLikeNode, arra
      */
     public function __debugInfo(): array
     {
-        return [
-            'name' => $this->getName()
-        ];
+        return [];
```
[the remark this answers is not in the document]
WalterWoshid: What exactly is `__debugInfo()` for?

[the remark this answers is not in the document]
lisachenko: This method is called by `var_dump()` and similar tools (including IDE too) to return an array with debug data of object.
Mainly it is used to hide complex internal data structure (eg low-level AST tree nodes, variables, etc) and return fake data to make object looks like it has one simple property, in this case - name. Better for logs, but worse for debugging )

See an example here: https://3v4l.org/FF7RY

lisachenko:
FYI, @WalterWoshid - you can also update https://github.com/goaop/parser-reflection/blob/master/.github/workflows/phpunit.yml#L18-L20 to trigger run only for PHP>=8.0 versions, eg 8.0, 8.1 instead existing lines

src/ReflectionClass.php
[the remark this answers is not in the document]
WalterWoshid: Ohhh I see, thanks for the information!

src/Traits/ReflectionClassLikeTrait.php
WalterWoshid: Re-added and fixed for PHP 8.0

https://github.com/goaop/parser-reflection/pull/122/commits/894cc827e0a0f14cc9426ab14cec2de37d67201c

WalterWoshid: Works with PHP 8.1 also

src/ReflectionClass.php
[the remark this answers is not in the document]
WalterWoshid: Re-added

https://github.com/goaop/parser-reflection/pull/122/commits/1585b64af3ce5d2ec0595d1d2a378fedc51ec635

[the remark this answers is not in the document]
WalterWoshid: Added description
https://github.com/goaop/parser-reflection/pull/122/commits/f8dd6098edbe29776fcbe5fa0d032f58b1b31a95

src/ReflectionMethod.php
WalterWoshid: https://github.com/goaop/parser-reflection/pull/122/commits/c331ba429e241fa8424fe59cccefdf3bff5f607e

src/ReflectionClass.php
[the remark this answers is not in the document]
WalterWoshid: When debugging the ReflectionClass constructor, sometimes the debugger crashes, because the ide or the debugger calls this method internally and the `getName` method in `__debugInfo` should always return a string, but throws an Exception because the constructor wasn't done yet. What should the `var_dump()` method return while inside the constructor? Just `null` values or and `Uninitialized` string?

src/Traits/ReflectionClassLikeTrait.php
WalterWoshid: Does it make sense to add more methods than the BaseReflection classes have? For example there is only `getAttributes`, but you could also add `getAttribute` and `hasAttribute`?

src/ReflectionClass.php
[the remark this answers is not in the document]
lisachenko: Yep, this can happen, indeed. 
Inside ctor, you can just return empty array or just return null for value of name (which is better and logical) to prevent exception throwing.

src/Traits/ReflectionClassLikeTrait.php
WalterWoshid: I see, then it will be like that :)

src/ReflectionProperty.php
WalterWoshid: Is the `ValueResolver\NodeExpressionResolver` ideal for the `__toString()` method for `ReflectionProperty`? I tried to re-add it, but it looks very different in PHP 8.0 and 8.1

WalterWoshid: Nevermind, I think I have to apply some custom logic

lisachenko: To be honest, this method is not used too much, so if you have troubles with it - then just simple skip it, it doesn't deserve too much time

WalterWoshid: Re-added, but not fully. It's not that important
https://github.com/goaop/parser-reflection/pull/122/commits/f8f982055a80da6707aab564167d9b024b314c00

composer.json
```diff
@@ -22,11 +22,11 @@
     }
   },
   "require": {
-    "php": ">=7.3",
+    "php": "^8.0",
     "nikic/php-parser": "^4.0"
```
lisachenko, line 26: For PHP>=8.0, this dependency should be bumped to at least ^4.10.5. Because only 4.10.0 supports PHP8 attributes

src/ReflectionFile.php
```diff
@@ -28,21 +29,21 @@ class ReflectionFile
...
-    protected $fileNamespaces;
+    protected array $fileNamespaces;
 
     /**
      * Top-level nodes for the file
      *
      * @var Node[]
      */
-    private $topLevelNodes;
+    private Node|array $topLevelNodes;
```
lisachenko, line 46: This variable can't hold `Node` type, as otherwise return type hint for the `getNodes()` method will be violated (it returns only arrays now)

src/ReflectionProperty.php
```diff
@@ -75,12 +78,144 @@ public function __construct(
...
+            $checkType   = true;
+
+            $default          = $this->getDefaultValue();
+            $defaultValueType = gettype($default);
+            $defaultNode      = $this->getNode()->default;
+            $declaringClass   = $this->getDeclaringClass();
+            $namespaceName    = $declaringClass->getNamespaceName();
+
+            // PHP >= 8.1 has changed the way how default values are represented
+            if (PHP_VERSION_ID >= 80100) {
```
lisachenko, line 102: This entire logic of parsing/analyzing default value should be only in one single place - `NodeExpressionResolver` inside `ValueResolver` namespace. It was created to work on top of AST + given context (eg class or method).

There can be same places for `ReflectionParameter`, `ReflectionClassConstant`, etc. This is why it will be better use AST tree + context in all places.

If you have troubles only with `__toString()` methods in all classes - let's skip them, as in reality such methods are not used very often. You can mark them as skipped for now to avoid spending a lot of time on preparing same output - this might be time-consuming process

WalterWoshid:
I see that some `ReflectionMethod` instances are just base instances, but they are called with `Go\ParserReflection\ReflectionClass->getMethods()`. Is it a bug from my refactoring's or is there no way to get a `Go\ParserReflection\ReflectionMethod`?

src/ReflectionProperty.php
WalterWoshid: I see, starting with that gave me a good look on the AST tree and how everything works.

src/ReflectionProperty.php
```diff
@@ -75,12 +78,144 @@ public function __construct(
...
+                    || $defaultNode instanceof MagicConst
+                ) {
+                    // __CLASS__ inside a trait returns __CLASS__
+                    // TODO: Not declaring class, but the class where the trait is used
+                    if ($declaringClass->isTrait()
+                        && $defaultNode instanceof MagicConst\Class_
+                    ) {
+                        $default = '__CLASS__';
+                        $quoteString = false;
+                    }
```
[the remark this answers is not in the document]
WalterWoshid: It's okay that way bro

composer.json
WalterWoshid: https://github.com/goaop/parser-reflection/pull/122/commits/e96f9bbaa0be7f40cda786257aa46d526558f6c4

src/ReflectionFile.php
WalterWoshid: https://github.com/goaop/parser-reflection/pull/122/commits/03feb2da4c23e65e5c8399c40621d1eefafa7a83

src/ReflectionMethod.php
```diff
@@ -47,16 +50,17 @@ class ReflectionMethod extends BaseReflectionMethod
      * @param string           $methodName      Name of the method
      * @param ?ClassMethod     $classMethodNode AST-node for method
      * @param ?ReflectionClass $declaringClass  Optional declaring class
+     *
+     * @noinspection PhpMissingParentConstructorInspection
      */
     public function __construct(
-        $className,
-        $methodName,
+        private string $className,
```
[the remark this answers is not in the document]
WalterWoshid: Nope

src/ReflectionMethod.php
```diff
@@ -47,16 +50,17 @@ class ReflectionMethod extends BaseReflectionMethod
...
+        string $methodName,
         ClassMethod $classMethodNode = null,
-        ReflectionClass $declaringClass = null
+        private ?ReflectionClass $declaringClass = null
```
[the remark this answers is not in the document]
WalterWoshid: Na-ah

src/Traits/ReflectionClassLikeTrait.php
WalterWoshid: WIP https://github.com/goaop/parser-reflection/pull/122/commits/b60ced001ace42fbb28d1a6c56c63f297d941793

WalterWoshid: Done! https://github.com/goaop/parser-reflection/pull/122/commits/4b2de97f0268f93dc766632dacbc48667d888253 with tests also https://github.com/goaop/parser-reflection/pull/122/commits/ae631c4a579285dbe4c7e48a3cd398c8ad7546c4

lisachenko:
@WalterWoshid any updates here so far? )

WalterWoshid:
Hello @lisachenko, I am currently working on my own project. Because I liked the idea of AOP so much, I am currently implementing it as my own library. I have broken it up into 2 libraries, a [PHP Code Transformer](https://github.com/okapi-web/php-code-transformer) and the AOP library, which is currently in work. I was heavily inspired from the system behind your library (StreamFilter, overloading the Composer ClassLoader, Kernel). I hope that's okay for you and if you want, I can give you some credit, but I'll probably do it anyway :)

WalterWoshid:
Closed, because this library is deprecated in favor of [BetterReflection](https://github.com/Roave/BetterReflection)

PR #124, opened 2024-01-08
[Step 1 Upgrade] Bump requirement to php 7.4

samsonasik:
@lisachenko here the 1st step 👍 , bump first to php 7.4


samsonasik:
`continuous-integration/travis-ci ` seems not active, or it no longer active? 

```
 continuous-integration/travis-ci Expected — Waiting for status to be reported 
```

I think that can be removed from required workflow

samsonasik:
Ready to merge 👍

PR #125, opened 2024-01-09
[Step 2 Upgrade] Bump requirement to php 8.0

samsonasik:
@lisachenko here the 2nd step 👍 , bump to php 8.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,11 +22,11 @@
     }
   },
   "require": {
-    "php": ">=7.4",
+    "php": ">=8.0",
     "nikic/php-parser": "^4.0"
   },
   "require-dev": {
-    "phpunit/phpunit": "^8.0"
+    "phpunit/phpunit": "^9.0"
   },
   "extra": {
     "branch-alias": {
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,8 @@
     "nikic/php-parser": "^4.0"
   },
   "require-dev": {
-    "phpunit/phpunit": "^9.0"
+    "phpunit/phpunit": "^9.0",
+    "tracy/tracy": "^2.10"
   },
   "extra": {
     "branch-alias": {
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
   },
   "extra": {
     "branch-alias": {
-      "dev-master": "3.x-dev"
+      "dev-master": "4.x-dev"
     }
   }
 }
--- a/src/ReflectionMethod.php
+++ b/src/ReflectionMethod.php
@@ -112,11 +112,10 @@ class ReflectionMethod extends BaseReflectionMethod
         }
 
         return sprintf(
-            "%sMethod [ <user%s%s%s>%s%s%s %s method %s ] {\n  @@ %s %d - %d{$paramFormat}{$returnFormat}\n}\n",
+            "%sMethod [ <user%s%s>%s%s%s %s method %s ] {\n  @@ %s %d - %d{$paramFormat}{$returnFormat}\n}\n",
             $this->getDocComment() ? $this->getDocComment() . "\n" : '',
             $prototype ? ", overwrites {$prototypeClass}, prototype {$prototypeClass}" : '',
             $this->isConstructor() ? ', ctor' : '',
-            $this->isDestructor() ? ', dtor' : '',
             $this->isFinal() ? ' final' : '',
             $this->isStatic() ? ' static' : '',
             $this->isAbstract() ? ' abstract' : '',
--- a/src/ReflectionMethod.php
+++ b/src/ReflectionMethod.php
@@ -203,17 +203,17 @@ class ReflectionMethod extends BaseReflectionMethod
     /**
      * {@inheritDoc}
      */
-    public function invoke($object, $args = null)
+    public function invoke(?object $object, mixed ...$args): mixed
     {
         $this->initializeInternalReflection();
 
-        return parent::invoke(...func_get_args());
+        return parent::invoke($object, ...$args);
     }
 
     /**
      * {@inheritDoc}
      */
-    public function invokeArgs($object, array $args)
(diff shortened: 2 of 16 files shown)


lisachenko:
@samsonasik There might be good parts which can be cherry-picked from the https://github.com/goaop/parser-reflection/pull/122

samsonasik:
I will check 👍

src/ReflectionType.php
```diff
@@ -93,7 +94,11 @@ public static function convertToDisplayType(BaseReflectionType $type)
         $displayType = ltrim($displayType, '\\');
 
         if ($type->allowsNull()) {
-            $displayType .= ' or NULL';
+            if (! $type instanceof ReflectionUnionType) {
+                $displayType = '?' . $displayType;
+            } else {
+                $displayType = '|null';
+            }
```
samsonasik, line 101: @lisachenko this is the BC break for display type, as original `ReflectionParameter::__toString()` on php 8.0 now uses `?` instead of `or NULL` for parameter type information.

tests/ReflectionTypeTest.php
```diff
@@ -85,14 +72,31 @@ public function testTypeConvertToDisplayTypeImplicitlyNullable()
...
+    {
+        $nativeClassRef = new \ReflectionClass('Go\\ParserReflection\\Stub\\ClassWithPhp80Features');
+        $nativeMethodRef = $nativeClassRef->getMethod('acceptsStringArrayDefaultToNull');
+        $this->assertEquals(\ReflectionMethod::class, get_class($nativeMethodRef));
+        $nativeParamRefArr = $nativeMethodRef->getParameters();
+        $this->assertCount(1, $nativeParamRefArr);
+        $this->assertEquals(\ReflectionParameter::class, get_class($nativeParamRefArr[0]));
+        $nativeTypeRef = $nativeParamRefArr[0]->getType();
+        $this->assertTrue($nativeTypeRef->allowsNull());
+        $this->assertEquals('array|string|null', $nativeTypeRef);
```
samsonasik, line 97: there is no `->getName()` on native `ReflectionUnionType`, so it just use its `__toString()` convert, got full type `array|string|null`

src/ReflectionProperty.php
```diff
@@ -113,11 +113,31 @@ public function __debugInfo(): array
...
+
+                $defaultValue = "'" . $defaultValue . "'";
+            }
+
+            if ($defaultValue === null) {
+                $defaultValue = "NULL";
+            }
+
+            $defaultValueDisplay = '= ' . $defaultValue;
+        }
```
samsonasik, line 134: @lisachenko this is for display property with default value display

src/ReflectionMethod.php
```diff
@@ -112,11 +112,10 @@ public function __toString()
         }
 
         return sprintf(
-            "%sMethod [ <user%s%s%s>%s%s%s %s method %s ] {\n  @@ %s %d - %d{$paramFormat}{$returnFormat}\n}\n",
+            "%sMethod [ <user%s%s>%s%s%s %s method %s ] {\n  @@ %s %d - %d{$paramFormat}{$returnFormat}\n}\n",
             $this->getDocComment() ? $this->getDocComment() . "\n" : '',
             $prototype ? ", overwrites {$prototypeClass}, prototype {$prototypeClass}" : '',
             $this->isConstructor() ? ', ctor' : '',
-            $this->isDestructor() ? ', dtor' : '',
```
samsonasik, line 119: @lisachenko `dtor` no longer displayed on php 8.0

samsonasik:
The current skipped `testGetModifiers()` seems needs separate PR:

https://github.com/goaop/parser-reflection/blob/1966e14afe27eb8ef2b7a71a73b6765f04f54e8d/tests/ReflectionClassTest.php#L31-L35

@lisachenko Ready to merge 👍

src/ReflectionType.php
lisachenko: I guess, it is ok here, we follow logic in the original ReflectionAPI. So, I'm ok to keep this in next major version for 8.x

src/ReflectionProperty.php
lisachenko: To be honest, this crazy export/_toString logic for some Reflection classes might be dropped - it is not needed directly by goaop/framework. Mostly, this was implemented for compatibility previously

src/ReflectionMethod.php
lisachenko: We are safe to drop this

src/ReflectionProperty.php
samsonasik: I am thinking that keeping map with native API is the first step, we can remove when everything compatible 👍

lisachenko:
> The current skipped `testGetModifiers()` seems needs separate PR:
> 
> https://github.com/samsonasik/parser-reflection/blob/f220c210ca2fc443b71466c864325d9ad0bc0470/tests/ReflectionClassTest.php#L31-L35
> 
> @lisachenko Ready to merge 👍

Awesome work! 👍  Let me check this today in the evening )

src/ReflectionProperty.php
lisachenko: Yes, this might be easier to achieve, indeed. 

Just wanted to highlight that these two methods can generate some troubles, but not used directly by the goaop/framework and we might just drop support for this method.

lisachenko:
Question here: do you plant to add PHP8 Attributes as separate independent PR?

composer.json
lisachenko, line 1: Could you please also bump the branch-alias to 4.x-dev as well?

samsonasik: updated 👍

samsonasik:
If you mean `ReturnTypeWillChange` attribute, that seems on php 8.1, the PR is step by step which this target is to make phpunit green first.

Otherwise, I may need failing test case for that to be implemented.

lisachenko (approved):
Approved, thanks

PR #127, opened 2024-01-15
[Step 4 Upgrade] Bump requirement to php 8.1

samsonasik:
@lisachenko here the 4th step 👍 , bump to php 8.1


src/Traits/ReflectionClassLikeTrait.php
```diff
@@ -519,7 +518,8 @@ function (array &$result, \ReflectionClass $instance, $isParent) {
     /**
      * {@inheritdoc}
      */
-    public function getProperty($name)
+    #[\ReturnTypeWillChange]
```
samsonasik, line 521: @lisachenko  this is needed due to original Reflection return `ReflectionProperty`, while it can return `ReflectionProperty|false`

src/Traits/ReflectionClassLikeTrait.php
```diff
@@ -345,9 +346,9 @@ function (array &$result, \ReflectionClass $instance) {
 
     /**
      * {@inheritdoc}
-     * @param string $name
      */
-    public function getMethod($name)
+    #[\ReturnTypeWillChange]
```
samsonasik, line 350: @lisachenko  this is needed due to original Reflection return `ReflectionMethod`, while it can return `ReflectionMethod|false`

lisachenko: Noted, thanks, I'm aware about this change.

src/Traits/ReflectionClassLikeTrait.php
lisachenko: Yes, confirmed, thank you

lisachenko:
Hello, @samsonasik! You are doing great work, I really appreciate your effort and your help on this 👍

lisachenko:
Please ping me, once it will be ready for merge and tests are green

src/ValueResolver/NodeExpressionResolver.php
```diff
@@ -278,6 +279,11 @@ protected function resolveExprClassConstFetch(Expr\ClassConstFetch $node)
         $refClass     = $this->fetchReflectionClass($classToReflect);
         $constantName = ($node->name instanceof Expr\Error) ? '' : $node->name->toString();
 
+        if ($node->class instanceof Name && $node->class->isSpecialClassName()) {
+            $this->isConstant = true;
+            return $node->class . '::' . $constantName;
+        }
```
samsonasik, line 285: @lisachenko special class name: `self::*, parent::*, static::*` is now printed as is.

samsonasik: I will look more why the behaviour is different between in ReflectionMethod vs ReflectionClass.

samsonasik: Looking directly on `Expr\ClassConstFetch` seems got invalid result, as it doesn't know what it come from (its parent node), I guess we can replace it with `PhpParser` Printer class instead.

samsonasik: using `ParentConnectingVisitor` seems can be solution, but need to verify `nodeLevel` https://github.com/goaop/parser-reflection/pull/127/commits/60194311af8fd1859a1c92a991be45290b055e9c

I will look more for append, eg, via concat with expanded test fixture

samsonasik: Note for different result start from php 8.1 https://3v4l.org/IoXSq

Screenshot 2024-01-18 at 23 19 14

samsonasik:
@lisachenko its 2 errors to go 👍 

```diff
There were 2 failures:

1) Go\ParserReflection\ReflectionParameterTest::testGeneralInfoGetters with data set "PHP5.5" ('/home/runner/work/parser-refl...55.php')
isDefaultValueAvailable() for parameter Go\ParserReflection\Stub\miscParameters:arrayParamWithDefault should be equal
Failed asserting that true is identical to false.

/home/runner/work/parser-reflection/parser-reflection/tests/ReflectionParameterTest.php:61

2) Go\ParserReflection\ReflectionParameterTest::testGeneralInfoGetters with data set "PHP7.0" ('/home/runner/work/parser-refl...70.php')
isDefaultValueAvailable() for parameter Go\ParserReflection\Stub\miscScalarParameters:acceptsFloat should be equal
Failed asserting that true is identical to false.

/home/runner/work/parser-reflection/parser-reflection/tests/ReflectionParameterTest.php:61
```

It somehow probably related with autoload, as manually create namespace before test make it work, but load the file and parse it cause invalid original `\ReflectionParameter` data.

I will look more...

src/ReflectionParameter.php
```diff
@@ -338,50 +349,66 @@ public function isCallable()
...
+            if ($key <= $this->parameterIndex) {
+                continue;
+            }
+
+            if (! $parameter->default instanceof Expr) {
+                return false;
+            }
+        }
+
+        return true;
```
samsonasik, line 377: @lisachenko this is needed, as since PHP 8.1, `isDefaultValueAvailable()` returns false if next parameter is required, see 

- https://github.com/php/php-src/issues/8090

samsonasik:

src/ReflectionParameter.php
```diff
@@ -132,23 +138,30 @@ public function __debugInfo(): array
...
-            /* @see https://3v4l.org/DJOEb for behaviour changes */
-            if (is_float($defaultValue) && fmod($defaultValue, 1.0) === 0.0) {
-                $defaultValue = (int)$defaultValue;
-            }
 
-            $defaultValue = str_replace('\\\\', '\\', var_export($defaultValue, true));
+            if ($this->parameterNode->default instanceof Array_) {
+                $this->parameterNode->default->setAttribute('kind', Array_::KIND_SHORT);
+                $printer = new Standard(['shortArraySyntax' => true]);
+                $defaultValue = $printer->prettyPrintExpr($this->parameterNode->default);
```
samsonasik, line 153: Note to myself: using Printer should pass the transformed default value on `___toString()`, eg; transformed array with magic const: `[___FUNCTION___ . "test"]` which should equivalent to `[namespaced\functiontest]`

but that may need separate PR

src/ReflectionParameter.php
lisachenko: Aha, I see... Thanks!

lisachenko:
>  implemented 🎉 @lisachenko Ready to merge 👍

Nothing to add here - great job @samsonasik! It's my pleasure to merge this good work )

lisachenko (approved):

PR #128, opened 2024-01-21
[Step 5 Upgrade] Bump requirement to php 8.2

samsonasik:
@lisachenko here the 5th step 👍 , bump to php 8.2

### DIFF
--- a/src/ReflectionParameter.php
+++ b/src/ReflectionParameter.php
@@ -320,6 +320,10 @@ class ReflectionParameter extends BaseReflectionParameter
             return null;
         }
 
+        if ($parameterType === 'iterable') {
+            $parameterType = 'Traversable|array';
+        }
+
         return new ReflectionNamedType($parameterType, $allowsNull, $isBuiltin);
     }
 
--- a/src/Traits/ReflectionFunctionLikeTrait.php
+++ b/src/Traits/ReflectionFunctionLikeTrait.php
@@ -198,6 +198,10 @@ trait ReflectionFunctionLikeTrait
             return null;
         }
 
+        if ($returnType === 'iterable') {
+            $returnType = 'Traversable|array';
+        }
+
         return new ReflectionNamedType($returnType, $isNullable, $isBuiltin);
     }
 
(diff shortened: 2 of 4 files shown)


samsonasik:

src/ReflectionParameter.php
```diff
@@ -320,6 +320,10 @@ public function getType(): ?\ReflectionType
             return null;
         }
 
+        if ($parameterType === 'iterable') {
+            $parameterType = 'Traversable|array';
+        }
```
samsonasik, line 325: @lisachenko this is needed as php 8.2 display `Traversable|array` for `iterable, see https://3v4l.org/3uVi3

src/Traits/ReflectionFunctionLikeTrait.php
```diff
@@ -198,6 +198,10 @@ public function getReturnType(): ?\ReflectionType
             return null;
         }
 
+        if ($returnType === 'iterable') {
+            $returnType = 'Traversable|array';
+        }
```
samsonasik, line 203: @lisachenko this is needed as php 8.2 display `Traversable|array` for `iterable, see https://3v4l.org/V1RCP

TomasVotruba:
Seems good to go :+1:

src/ReflectionParameter.php
lisachenko: Oh, nice )) Haven't been aware about this 👍

lisachenko:
Here we go! 🕺

lisachenko (approved):

lisachenko:
Btw, do you know where this annoying travis-ci check comes from? I've checked every tab, removed travis-ci from everywhere, but still this check is here...

lisachenko:
Congrats, @samsonasik! 

The only missing final part here, for parser-reflection now is ReflectionAttribute + relevant attributes API, which can be ported from https://github.com/goaop/parser-reflection/pull/122/files#diff-538c6d3a73a6c88b7717c7483cefab2a2f7c9ff32c5267a96ed2d6cbaaccc0d3

samsonasik:
@lisachenko thank you for merging, I will look into the attribute part. I don't know about travis CI part as I don't have admin access on this repo, probably activated in API request somewhere in `/settings/installations` `settings/hooks` need to be disabled.

PR #130, opened 2024-01-24
[Step 6 Upgrade] Add PHP 8+ Attributes support

samsonasik:
@lisachenko here the 6th step 👍 , to add attributes support.


lisachenko:
Hello, @samsonasik! 

Do you plan to add actual `ReflectionAttribute` implementation later, so ReflectionMethod->getAttributes(), ReflectionClass->getAttributes() and similar from AST

samsonasik:
Hi @lisachenko , yes, I will try it in this PR

src/ReflectionAttribute.php
```diff
@@ -0,0 +1,43 @@
...
+use PhpParser\Node;
+use Reflector;
+
+class ReflectionAttribute extends BaseReflectionAttribute
+{
+    use InternalPropertiesEmulationTrait;
+
+    public function __construct(
+        string                  $attributeName,
+        private ?Node\Attribute $attributeNode = null,
```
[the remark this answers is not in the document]
lisachenko: Nevermind, it is just old Nitpick-CI set of rules )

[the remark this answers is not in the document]
samsonasik: can we disable Nitpick-CI and scrutinizer service? It probably better to use coding standard check on github action itself, that can be run separately to fix it

src/ReflectionEngine.php
```diff
@@ -80,6 +81,7 @@ public static function init(LocatorInterface $locator): void
         self::$traverser = $traverser = new NodeTraverser();
         $traverser->addVisitor(new NameResolver());
         $traverser->addVisitor(new RootNamespaceNormalizer());
+        $traverser->addVisitor(new ParentConnectingVisitor());
```
samsonasik, line 84: this is needed to go up from `PropertyProperty` to get `Property`, which has `attrGroups`, via `parent` attribute.

samsonasik:

lisachenko:
Hello, @samsonasik! This is awesome ) Going to test/review it either today in the evening or during weekend. Great job! 

I'm very grateful for your contribution!

lisachenko:
Hello, @samsonasik, I'm testing now this PR and it looks like the `getAttributes()` method inside the `AttributeResolverTrait` is triggering the class load and doesn't work with raw AST tree. 

Unfortunately, this will not work for the goaop/framework, as it analyzes source code during the autoload process and attempt to call native reflection here will just silently fail with error that class has not been loaded yet. That's why this entire library has been created - to provide reflection just from the raw AST nodes.

So, proper implementation of the `getAttributes()` method should take the underlying AST node (eg, parameterNode, functionLikeNode, classLikeNode, etc)  and use relevant `attrGroups` (AttributeGroup array) from such nodes (classLikeNode->attrGroups) to pass this information to the constructor of Go\ParserReflection\ReflectionAttribute (each Attribute node AST instance from the attrGroups should be wrapped with own instance of corresponding ReflectionAttribute)

To test this - create a random class with attribute and put before or after definition of such class special "echo" statement, like this:

```php
use JetBrains\PhpStorm\ArrayShape;
use JetBrains\PhpStorm\Deprecated;

#[Deprecated]
class FooBar
{
    #[ArrayShape([
        'token' => 'string',
        'code' => 'integer'
    ])]
    public $foo;
}
echo "Loaded - Runtime Error!"; // Might be replaced with exception
```

And then try to reflect this class via parser reflection to read these attributes, you should not see this echoed message.

```php

use Go\ParserReflection\Locator\ComposerLocator;
use Go\ParserReflection\ReflectionEngine;

include __DIR__ . '/vendor/autoload.php';

ReflectionEngine::init(new ComposerLocator());

$parsedClass = new \Go\ParserReflection\ReflectionClass(\Go\ParserReflection\FooBar::class);
var_dump($parsedClass->getAttributes());
```

Hope, this information will help you to understand the issue inside current PR and fix it. Thank you again for such huge work, it's almost finished )

src/Traits/ReflectionFunctionLikeTrait.php
```diff
@@ -219,6 +219,12 @@ public function getShortName(): string
 
     public function getStartLine(): int
     {
+        // php parser detect same line for has attrGroups
+        // attrGroup is mostly on next line
+        if ($this->functionLikeNode->attrGroups !== []) {
+            return $this->functionLikeNode->getAttribute('startLine') + count($this->functionLikeNode->attrGroups);
```
lisachenko, line 225: Will it work for multi-line attributes? Like this:
```php
#[Deprecated]
class FooBar
{
    #[ArrayShape([
        'token' => 'string',
        'code' => 'integer'
    ])]
    public $foo;
}
```

I guess, this applies to all relevant getStartLine() methods too. Not sure, but check should be done

src/Traits/AttributeResolverTrait.php
```diff
@@ -0,0 +1,36 @@
...
+use Go\ParserReflection\ReflectionAttribute;
+
+trait AttributeResolverTrait
+{
+    /**
+     * @return ReflectionAttribute[]
+     */
+    public function getAttributes(?string $name = null, int $flags = 0): array
+    {
+        $this->__initialize();
```
lisachenko, line 24: Initialization is not allowed here, as it will trigger the load of original class and for the `goaop/framework` this won't work, as we will be already inside class loading process, so fatal error will occur that class has not been loaded yet.

So, this call should be disabled and all attributes should be crafted from the corresponding AST nodes here. 

To have an idea - it might be good to have a look at `\Go\ParserReflection\ReflectionMethod::collectFromClassNode` and create something similar, eg `\Go\ParserReflection\ReflectionAttribute::collectFromAttrGroup(AttrGroup $attrGroup, ReflectionClass|ReflectionMethod|ReflectionParameter|<anything else containing attribute> $reflector): ReflectionAttribute[]

lisachenko (changes requested):
So, main issue here is to make `getAttributes()` method working on pure AST without delegation to the parent Reflection API.

src/Traits/ReflectionFunctionLikeTrait.php
samsonasik: That's an issue with php-parser, that can't detect line for attribute, as always fall for the stmt that hold it, so attrGroup or attr line = stmt line.

samsonasik: I will look if that possible to tweak

src/Traits/AttributeResolverTrait.php
samsonasik: `__initialize()` were not always executed before, that's why I added the new method `__initialize()` in some reflections in this PR, otherwise we got fatal error on `getAttributes()` call.

 I will check if init can be called early in the reflection itself.

src/Traits/ReflectionFunctionLikeTrait.php
samsonasik: The solution is to use original start line from native reflection itself when attribute exists, see https://github.com/goaop/parser-reflection/pull/130/commits/a06b0c77f9e4fbae37d516203042a3ee9f71284c

samsonasik:
@lisachenko I added failing test case with stub with throw RuntimeException inside it, do you mean unit test should not show error even it has `RuntimeException` inside stub https://github.com/goaop/parser-reflection/pull/130/commits/031c9dbcfd71dcf5c3fab778730c1ae26769e2b1 : 

```
There was 1 error:

1) Go\ParserReflection\Locator\ComposerLocatorTest::testLocateClassWithAttributes
RuntimeException: test
```

If I remove the `$this->__initialize()` usage, it got error:

```
There was 1 error:

1) Go\ParserReflection\Locator\ComposerLocatorTest::testLocateClassWithAttributes
Error: Internal error: Failed to retrieve the reflection object

/Users/samsonasik/www/parser-reflection/src/Traits/AttributeResolverTrait.php:28
```

I tried cheking from the node from its reflector and got segmentation fault:

```
Runtime:       PHP 8.3.0
Configuration: /Users/samsonasik/www/parser-reflection/phpunit.xml.dist

.[1]    7644 segmentation fault  vendor/bin/phpunit tests/Locator/ComposerLocatorTest.php
```

```php
    public function getAttributes(?string $name = null, int $flags = 0): array
    {
        $node = $this->getNode();
        $attributes = [];

        foreach ($node->attrGroups as $attrGroup) {
            foreach ($attrGroup->attrs as $attr) {
                if ($name === null) {
                    $attributes[] = new ReflectionAttribute($attr->name->toString(), $this);

                    continue;
                }

                if ($name !== $attr->name->toString()) {
                    continue;
                }

                $attributes[] = new ReflectionAttribute($name, $this);
            }
        }

        return $attributes;
    }
```

samsonasik:
@lisachenko I added handling get Attributes by AST https://github.com/goaop/parser-reflection/pull/130/commits/9a2bb99af4fe2a7f4f222035acbaa447fbed5c35 , not sure yet on `getTarget()` tho, as it not documented in php manual https://www.php.net/manual/en/reflectionattribute.gettarget.php

samsonasik:
@lisachenko I updated to get Line via Node which go to end of `attrGroups` line and add + 1 (mostly next line of attrGroups), and throw the `RuntimeException` on `getTarget()` and `newInstance()` as it require the real ReflectionAttribute https://github.com/goaop/parser-reflection/pull/130/commits/b5674732f02a1a78b64acd78f3fdf15e22210cc2

lisachenko:
Wow! This was fast! ))

I've tested it for the class and method - it works!! 🚀  

There is only gotcha with properties (would be nice to add a test for this) - for ReflectionProperty attributes are stored not inside `PropertyProperty::$attrGroups`, but inside type node (instance of `Property` AST node, which might be accesses via `getTypeNode()` getter.

```
PHP Deprecated:  Return type of Go\ParserReflection\ReflectionProperty::getValue($object = null) should either be compatible with ReflectionProperty::getValue(?object $object = null): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in ./parser-reflection/src/ReflectionProperty.php on line 196
PHP Warning:  Undefined property: PhpParser\Node\Stmt\PropertyProperty::$attrGroups in ./parser-reflection/src/Traits/AttributeResolverTrait.php on line 29
PHP Warning:  foreach() argument must be of type array|object, null given in ./parser-reflection/src/Traits/AttributeResolverTrait.php on line 29
```

For ReflectionParameter, attribute also generates this deprecation notice:
`PHP Deprecated:  Return type of Go\ParserReflection\ReflectionParameter::getDefaultValue() should either be compatible with ReflectionParameter::getDefaultValue(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in ./parser-reflection/src/ReflectionParameter.php on line 265
`

lisachenko:
> @lisachenko I added handling get Attributes by AST [9a2bb99](https://github.com/goaop/parser-reflection/commit/9a2bb99af4fe2a7f4f222035acbaa447fbed5c35) , not sure yet on `getTarget()` tho, as it not documented in php manual https://www.php.net/manual/en/reflectionattribute.gettarget.php

We can skip this function implementation, or just delegate to the parent call. I guess, it should load the attribute class into memory, then check the bitmask for which type of items this attribute should be applied. We don't need this functionality now.

samsonasik:
@lisachenko sure, resolved at eefee84ba4781640681b57d66759ef143f2af204 on getting attrGroups from`Property` 👍 , it should be ready now 👍

samsonasik:
I also updated `ReflectionParameter::getDefaultValue()` return type https://github.com/goaop/parser-reflection/pull/130/commits/799e69d32c5f365da030f1b4ce16338460918de7

lisachenko:
@samsonasik could you please check X/Twitter private messages BTW? ) Need some info

samsonasik:
@lisachenko yes, replied 👍

samsonasik:

lisachenko (approved):

lisachenko:
Merged, thanks!

PR #133, opened 2024-03-18
New 4.0 version implementation for PHP>=8.2

lisachenko:


marcovtwout:
@lisachenko With all the work you have performed, this line can probably be removed? https://github.com/goaop/parser-reflection/blob/master/README.md?plain=1#L3

lisachenko:
@marcovtwout nice comment here, but as you can see, current project is not actively maintained, mostly for my own personal needs. So, I think, it is more honest to report about this in readme.

For example, I haven't time for PHP8.4 features ( And new version is already here...

PR #180, opened 2026-04-27
Method isSubclassOf() should also look at interfaces

### LINKED ISSUE #72: implementsInterface() isSubclassOf() method inconsistancies in ReflectionClass

lisachenko:


lisachenko:
@copilot rebase or merge master and check if this PR fixes the original problem in #72

lisachenko:
@copilot build is failing, please fix

Issue #223, opened 2026-08-25
[PHP8.6] Readonly property default values


lisachenko:
Part of epic #219.

## Feature

PHP 8.6 (Readonly Property Defaults RFC) lifts the restriction that `readonly` properties cannot declare default values:

```php
final class CreateBooksTable implements MigratesUp
{
    public readonly string $name = '2026-01-01_create_books_table';
}
```

## Scope

- Ensure parsing a readonly property **with a default value** works end-to-end (PHP-Parser already parses the syntax; previously it was a compile error so the combination is untested here).
- Verify/fix the affected API surface in `src/ReflectionProperty.php` and `Traits/ReflectionClassLikeTrait.php`:
  - `hasDefaultValue()` / `getDefaultValue()` on readonly properties
  - `ReflectionClass::getDefaultProperties()` including readonly defaults
  - interaction with promoted readonly parameters and typed properties
- Check native 8.6 semantics for whether such a property still reports uninitialized-related behavior consistently and mirror it.
- Stub file with readonly defaults (plain, typed, const-expression defaults resolved by `NodeExpressionResolver`, class constants, enum cases as defaults) + parity tests guarded for 8.6 runtime, direct assertions on 8.5.

## Acceptance criteria

- Static results match native reflection on 8.6 for the whole stub matrix.
- PHPUnit green on 8.5, PHPStan level 10 clean.

PR #228, opened 2026-08-25
[PHP 8.6] Readonly property defaults — static behavior + upstream status

lisachenko:
Refs #223

## Upstream status first: the feature is NOT in `php-8.6.0beta1`

Before writing anything, the assumption behind #223 was verified against a real 8.6 binary:

```
$ php8.6 -r 'class A { public readonly string $x = "d"; }'
Fatal error: Readonly property A::$x cannot have default value
```
(PHP 8.6.0beta1, built from the `php-8.6.0beta1` tag.)

What the sources say:

| Fact | Source |
| --- | --- |
| RFC *Readonly Property Defaults* is **accepted** (24 yes / 0 no / 5 abstain, voting 2026-07-24 → 2026-08-06), status *Implemented*, target **PHP 8.6** | https://wiki.php.net/rfc/readonly_property_defaults |
| Implementation `[RFC] Allow Readonly Property Defaults (GH-22588)`, commit `a2640ae`, **merged into `master` on 2026-08-11** | https://github.com/php/php-src/pull/22588 |
| `UPGRADING` on `master` (still the 8.6 line — no `PHP-8.6` branch exists yet) lists it under *PHP 8.6 → New Features → Core*: “Readonly properties may now declare default values.” | https://github.com/php/php-src/blob/master/UPGRADING |
| The `php-8.6.0beta1` tag was cut **2026-08-11** and does not contain the commit (no `beta2` tag yet) | https://github.com/php/php-src/tags |

So the feature is genuinely coming in 8.6, it simply missed the beta1 snapshot by hours and should appear in the next beta/RC. Nothing here is blocked by that, but **the final native semantics can only be re-verified once a build ships it** — please decide whether to hold #223 until then.

Consequence for this PR: native parity assertions are impossible today, because no available runtime can even compile such a class. Everything asserted here is the *static* side, which PHP-Parser handles regardless of runtime support.

## What was found

Static reflection already handles readonly defaults correctly — the readonly modifier turned out to be orthogonal to default-value resolution:

* `hasDefaultValue()` / `getDefaultValue()` work for readonly properties with scalar, `null`, array, constant-expression and class-constant defaults, at every visibility, in a `readonly class`, and via traits/inheritance.
* `ReflectionClass::getDefaultProperties()` includes readonly defaults and correctly omits readonly properties without one.
* Promoted readonly parameters keep native semantics: the property reports no default (see bugs.php.net/81386), the parameter reports it.

One real defect surfaced while probing that surface:

```php
$property->getDefaultValueExpression();
// Error: Typed property Go\ParserReflection\ReflectionProperty::$defaultValueConstExpr
//        must not be accessed before initialization
```

`$defaultValueConstExpr` / `$defaultValueConstantName` were assigned only inside the “has a default value” branch of the constructor, so the getter fataled for any typed declaration **without** a default — e.g. `public readonly string $noDefault;`. The same latent fatal existed in `ReflectionParameter` (plus an uninitialized `$isDefaultValueConstExpr`).

## Changes

* `src/ReflectionProperty.php`, `src/ReflectionParameter.php` — initialize the default-value expression backing fields (`= null` / `= false`). Three-line, behavior-preserving fix; nothing else in the readonly path needed changing.
* `tests/Stub/FileWithReadonlyDefaults86.php` — new parse-only stub covering the full matrix from #223: scalar / `null` / const-expression / class-constant / enum-case / array defaults, all visibilities, `private(set) readonly`, readonly property without default, `readonly class`, trait + inheritance, promoted readonly parameters. It is never included or autoloaded (PSR-4 cannot resolve it, same convention as `FileWithFinalPromoted85.php`).
* `tests/ReadonlyPropertyDefaultsTest.php` — 13 tests documenting the static behavior, valid on 8.5 and 8.6 today.

### About the guarded parity test

`testNativeParityForReadonlyDefaults()` compares parsed vs native for the whole stub, but is gated on a runtime feature probe. The rejection is an **uncatchable compile-time fatal error**, so `eval()` in a `try/catch` cannot probe it — it kills the process. The probe therefore runs `PHP_BINARY -n -r 'class … { public readonly int $probe = 1; }'` in a short-lived child process and checks the exit code, caching the result. It skips everywhere today, and will activate by itself on the first runtime that accepts the syntax. Marked with a `TODO` to re-verify expectations against final native semantics at that point.

### Known limitation, deliberately not asserted as correct

An enum case used as a default (`public readonly Suit $suit = Suit::Spades;`) cannot be materialized statically while the enum is not loaded — `NodeExpressionResolver` falls back to a parsed `ReflectionClass`, whose `getConstants()` does not expose enum cases, so the value comes back as `false` instead of the case instance. This is a **pre-existing, general** gap (reproducible with a plain non-readonly property, unrelated to `readonly`), and producing a real enum instance is impossible without loading the enum. The test therefore asserts only `hasDefaultValue()` and the rendered expression, and does not enshrine the `false`. Worth a separate issue if parity there is wanted.

Also noted while probing, left untouched as out of scope: `ReflectionProperty::isInitialized()` with no argument returns `hasDefaultValue()`, whereas native throws `TypeError: Argument #1 ($object) must be provided for instance properties`; and `__toString()` renders const-expression defaults as the source expression (`self::PREFIX . 'books'`) where native renders the evaluated value (`'migration_books'`).

## Local results

| Check | Result |
| --- | --- |
| `vendor/bin/phpunit` (PHP 8.5.9) | OK — 13756 tests, 15481 assertions, 134 skipped, 2 incomplete |
| `php8.6 vendor/bin/phpunit` (PHP 8.6.0beta1) | OK — 13756 tests, 15481 assertions, 134 skipped, 3 incomplete |
| `vendor/bin/phpstan analyse src --no-progress` (level 10) | `[OK] No errors` |

Baseline on `master` was 13743 tests; the 13 new tests account for the difference, 1 of them skipped (the parity test) on both runtimes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Sy8BcM8ivUEpu8uVm7wADn

---
_Generated by [Claude Code](https://claude.ai/code/session_01Sy8BcM8ivUEpu8uVm7wADn)_

### DIFF
--- a/src/ReflectionParameter.php
+++ b/src/ReflectionParameter.php
@@ -62,7 +62,7 @@ final class ReflectionParameter extends BaseReflectionParameter implements NodeA
      *
      * @see $isDefaultValueConstant
      */
-    private ?string $defaultValueConstantName;
+    private ?string $defaultValueConstantName = null;
 
     /**
      * Index of parameter in the list
@@ -74,9 +74,9 @@ final class ReflectionParameter extends BaseReflectionParameter implements NodeA
      */
     private Param $parameterNode;
 
-    private bool $isDefaultValueConstExpr;
+    private bool $isDefaultValueConstExpr = false;
 
-    private ?string $defaultValueConstExpr;
+    private ?string $defaultValueConstExpr = null;
 
     private \ReflectionUnionType|\ReflectionNamedType|\ReflectionIntersectionType|null $type = null;
 
--- a/src/ReflectionProperty.php
+++ b/src/ReflectionProperty.php
@@ -65,11 +65,11 @@ final class ReflectionProperty extends BaseReflectionProperty implements NodeAwa
 
     private bool $isDefaultValueConstant = false;
 
-    private ?string $defaultValueConstantName;
+    private ?string $defaultValueConstantName = null;
 
     private bool $isDefaultValueConstExpr = false;
 
-    private ?string $defaultValueConstExpr;
+    private ?string $defaultValueConstExpr = null;
 
     /**
      * Initializes a reflection for the property
--- a/tests/ReadonlyPropertyDefaultsTest.php
+++ b/tests/ReadonlyPropertyDefaultsTest.php
@@ -1,0 +1,353 @@
+<?php
+declare(strict_types=1);
+/**
+ * Parser Reflection API
+ *
+ * @copyright Copyright 2025, Lisachenko Alexander <lisachenko.it@gmail.com>
+ *
+ * This source file is subject to the license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Go\ParserReflection;
+
+use Go\ParserReflection\Locator\CallableLocator;
+use Go\ParserReflection\Locator\ComposerLocator;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Documents how static reflection handles readonly properties that declare a default value.
+ *
+ * The "Readonly Property Defaults" RFC (https://wiki.php.net/rfc/readonly_property_defaults) was
+ * accepted for PHP 8.6, but its implementation landed in php-src after the `php-8.6.0beta1` tag,
+ * therefore no released runtime accepts this syntax yet: even 8.6.0beta1 aborts with
+ * "Readonly property X::$y cannot have default value" at compile time.
+ *
+ * PHP-Parser parses the syntax regardless, so the whole matrix can be verified statically today.
+ * The stub file must therefore never be included or autoloaded, it is only parsed. Assertions that
+ * require executing such a class are collected in the single parity test at the bottom, which stays
+ * skipped until the runtime really accepts readonly defaults (detected by a sub-process probe,
+ * because the failure is an uncatchable compile-time fatal error and cannot be probed via eval()).
+ */
+class ReadonlyPropertyDefaultsTest extends TestCase
+{
+    public const STUB_FILE = '/Stub/FileWithReadonlyDefaults86.php';
+
+    public const STUB_NAMESPACE = 'Go\ParserReflection\Stub\\';
+
+    public const STUB_CLASS = self::STUB_NAMESPACE . 'ClassWithReadonlyDefaults86';
+
+    public const READONLY_CLASS = self::STUB_NAMESPACE . 'ReadonlyClassWithDefaults86';
+
+    public const PROMOTED_CLASS = self::STUB_NAMESPACE . 'ClassWithPromotedReadonlyDefaults86';
+
+    public const CHILD_CLASS = self::STUB_NAMESPACE . 'ChildWithReadonlyDefaults86';
+
+    private string $stubFileName;
+
+    protected function setUp(): void
+    {
+        $resolvedFileName = stream_resolve_include_path(__DIR__ . self::STUB_FILE);
+        $this->assertIsString($resolvedFileName, 'PHP 8.6 readonly-defaults stub file should be available');
+
+        $this->stubFileName = $resolvedFileName;
+    }
+
+    protected function tearDown(): void
+    {
+        // Restores the default locator for the following tests, because this one replaces it
+        ReflectionEngine::init(new ComposerLocator());
+    }
+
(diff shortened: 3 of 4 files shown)


lisachenko:
Let's postpone here the implementation, until new version of 8.6 will be released. @claude you can unsubscribe from this PR, schedule server callback in 1 month to recheck if this fixed in PHP8.6 already or not, reply in thread

lisachenko:
Understood — postponing this until a PHP 8.6 build that actually ships readonly property defaults (the RFC implementation merged into php-src on 2026-08-11, right after the beta1 tag was cut, so beta2/RC should carry it).

I've scheduled a recheck for **2026-09-25**: it will probe whether a released 8.6 build accepts `public readonly int $x = 1;`, re-validate this PR's test matrix against it, and report back in this thread either way. I've also unsubscribed from this PR's activity until then, as requested.

For reference, the branch is already up to date with `master` and fully CI-green; the parity test in `ReadonlyPropertyDefaultsTest` self-activates via a runtime feature probe, so it needs no code change once a supporting build lands — only re-verification of the final native semantics.

---
_Generated by [Claude Code](https://claude.ai/code)_

lisachenko:
@claude php8.6 beta 2 has been released, should we recheck this now?

lisachenko:
are there any updates here with newer PHP builds, @claude ?

lisachenko:
On it now — apologies for the silence on the two pings: I was unsubscribed from this PR per the postponement and only saw them today (re-subscribed).

One correction to the premise: **there is no `php-8.6.0beta2` tag** — php-src went straight from beta1 to RC1/RC2. I'm building **PHP 8.6.0RC2** from source in the work environment right now (both RCs postdate the readonly-defaults merge of 2026-08-11, so RC2 should ship the feature). Once the build lands I'll probe `public readonly int $x = 1;`, bring the branch up to date with `master`, run the full test matrix — the parity test here self-activates on a supporting runtime — re-verify the final native semantics against the RC, and report results in this thread.

---
_Generated by [Claude Code](https://claude.ai/code)_

lisachenko:
**Done — readonly property defaults verified against PHP 8.6.0RC2. The hold reason is resolved and this PR is ready for review.**

I built `php-8.6.0RC2` from source in the work environment (there is no beta2 tag — the line went beta1 → RC1 → RC2) and re-validated the whole matrix. `public readonly string $x = 'def';` compiles and behaves as this PR's static reflection predicted: **all 13 readonly-defaults tests pass with the native parity test now active** (it self-activated via the runtime probe, exactly as designed — no changes to the parity expectations were needed).

Two 8.6 runtime changes surfaced during verification, handled in the new commit (`3f6619a`, on top of a fresh `master` merge):

1. **PHP 8.6 deprecates `ReflectionProperty::getDefaultValue()` on a property without a default** ("use `hasDefaultValue()` to check"). The parity test now compares values only for properties that have one, keeping the suite deprecation-clean. Note for a possible follow-up: the parsed `getDefaultValue()` still silently returns `null` in that case — mirroring the native deprecation in the library is a design decision I've left to you.
2. **Native `ReflectionClassConstant::__toString()` renders booleans as `true`/`false` since 8.6** (string-cast `1`/`` before). The parsed rendering now follows the runtime version. The `PHP_VERSION_ID >= 80600` guard needed a documented `phpstan.neon` ignore because the installed PHPStan caps its PHP version model at 8.5.x and folds the branch to "always false" (its config rejects `phpVersion.max` ≥ 80600 outright).

Validation on the pushed head:
| Check | Result |
| --- | --- |
| `ReadonlyPropertyDefaultsTest` on PHP 8.6.0RC2 | OK — 13 tests, 107 assertions, parity active, 0 deprecations |
| Full suite on PHP 8.6.0RC2 (`--fail-on-deprecation`) | OK — 15054 tests, 30492 assertions, 0 failures |
| Full suite on PHP 8.5.9 | OK — 13823 tests, 0 failures |
| PHPStan level 10 | `[OK] No errors` |

The CI 8.6 experimental job installs the latest 8.6 snapshot via setup-php, so the parity test should activate there as well. Merge is yours.

---
_Generated by [Claude Code](https://claude.ai/code)_

